PR #57, opened 2019-09-20
Remove boilerplate

bifurcation:
This is a clean-up PR that simplifies several things, especially around TLS serialization

* TLS serialization operators `<<` and `>>` can now be auto-generated based on a list of fields (`TLS_SERIALIZABLE(...)`)
* For classes that are TLS serializable, as above, new template `==` and `!=` operators apply
* Enum classes are now automatically TLS serialized
* GroupOperation now uses `std::variant` instead of a home-grown union approach
* `typedef` replaced by `using`

Submitting now to get CI and make sure this all works on Linux.  Might add more later.


bifurcation:
Overcome by events.

PR #59, opened 2019-10-25
Enable updates to DH keys

bifurcation:
This PR implements the core of the "updateable PKE" scheme described by Joël Alwen on the mailing list.  It could be the basis for implementing re-randomized TreeKEM.


bifurcation:
Closing because this PR is way out of date and the UPKE proposal was never adopted by the WG.

Issue #63, opened 2019-12-04
Re-enable session interop tests


bifurcation:
Something in the work for draft-08 compatibility broke the session interop tests, so I disabled them for the moment.  They should be re-enabled ASAP.

PR #64, opened 2019-12-04
First parts of draft-08

bifurcation:
This PR does the hardest parts of draft-08 migration, namely:

* Multi-recipient Welcome, and
* Proposal / Commit

The Proposal, Commit, and Welcome structures do not match what is in the spec, but this is because the spec is not workable.  So we'll need to fix the spec.

There are several other parts to get in order before draft-08 compatibility will be complete.  Filed several issues.


bifurcation:
Fixes #63

PR #84, opened 2020-06-24
Scale down tree math tests

bifurcation:
Using 255 leaves in the tree math tests was resulting in huge test vector files and tests taking forever (like 20sec per test, on a MacBook).  This PR just scales down the test size to make it go faster.  I don't think this degrades the testing value.


bifurcation:
Other changes are just clang-format.

PR #85, opened 2020-06-24
Remove private keys from structs

bifurcation:
This way, KeyPackage and Credential are purely public structs.

### DIFF
--- a/src/state.cpp
+++ b/src/state.cpp
@@ -22,51 +22,27 @@ State::State(bytes group_id,
 }
 
 // Initialize a group from a Welcome
-State::State(const std::vector<KeyPackage>& my_key_packages,
+State::State(const HPKEPrivateKey& init_priv,
+             const SignaturePrivateKey& sig_priv,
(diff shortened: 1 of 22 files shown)


src/state.cpp
```diff
@@ -9,65 +9,42 @@ namespace mls {
...
-    throw InvalidParameterError("Unable to decrypt Welcome message");
+  if (kp.cipher_suite != welcome.cipher_suite) {
+    throw InvalidParameterError("Ciphersuite mismatch");
   }
 
+  // Decrypt the GroupSecrets
+  auto secrets_ct = welcome.secrets[kpi].encrypted_group_secrets;
+  auto secrets_data = init_priv.decrypt(kp.cipher_suite, {}, secrets_ct);
+  auto secrets = tls::get<GroupSecrets>(secrets_data);
+
```
suhasHere, line 47: nit s/secrets/group_secrets

bifurcation: I was going for brevity :)

PR #87, opened 2020-06-25
RatchetTree -> TreeKEMPublicKey/PrivateKey

bifurcation:
The idea here is the same as https://github.com/cisco/go-mls/pull/71.  It brings along with it a lot of the other logic involved in upgrading to draft-09, including DirectPath updates and eventually extensions.  Still TODO:

- [x] Tests for TreeKEMPrivateKey
- [x] Tests for TreeKEMPublicKey
- [x] Integrate with State and remove 
- [x] Delete old RatchetTree code
- [x] ~Add path hashes to DirectPath~ (follow on bc of dependency on extensions)


include/messages.h
```diff
@@ -117,24 +101,24 @@ struct GroupInfo {
...
                    confirmed_transcript_hash,
                    interim_transcript_hash,
-                   path,
                    confirmation,
                    signer_index,
                    signature);
 };
 
+enum class NodeType : uint8_t;
+
```
suhasHere, line 112: am i reading it wrong, or do we have it defined twice (around line 30)

bifurcation: This is a forward declaration.

PR #88, opened 2020-06-27
Move from `tls::` types to type traits

bifurcation:
The current TLS syntax serialization code relies on subclassing STL container types, which is [not safe](https://stackoverflow.com/questions/4353203/thou-shalt-not-inherit-from-stdvector).  This has led to some odd errors in the development of #87, and I suspect it is responsible for some memory leakage that has been observe elsewhere (due to vectors not being properly cleaned up).

This PR replaces custom types with external type traits for type members.  These traits are specified using a `TLS_TRAITS` macro, parallel to the `TLS_SERIALIZABLE` macro.  For example:

```
// BEFORE
struct Foo {
  uint16_t a;
  Baz b;
  tls::vector<Bar, 3> c;

  TLS_SERIALIZABLE(a, b, c);
};

// AFTER

struct Foo {
  uint16_t a;
  Baz b;
  std::vector<Bar> c;

  TLS_SERIALIZABLE(a, b, c);
  TLS_TRAITS( tls::pass{}, tls::pass{}, tls::vector_trait<3>{} );
};
```

Together with the fact that several of the specialized types are not used, we should be able to use this construct to move basically all of the non-trivial serialization logic to traits.

The remaining things to do here are:

- [x] Remove unused specialized types
- [x] Create a trait to handle variant
- [x] Create a trait to handle optional (if necessary)
- [x] Migrate existing structs to use traits as required

### DIFF
--- a/include/messages.h
+++ b/include/messages.h
@@ -25,9 +25,10 @@ enum class ProtocolVersion : uint8_t
 struct RatchetNode
 {
   HPKEPublicKey public_key;
-  tls::vector<HPKECiphertext, 2> node_secrets;
+  std::vector<HPKECiphertext> node_secrets;
 
   TLS_SERIALIZABLE(public_key, node_secrets);
+  TLS_TRAITS(tls::pass{}, tls::vector_trait<2>{});
 };
 
 // struct {
@@ -35,9 +36,10 @@ struct RatchetNode
 // } DirectPath;
 struct DirectPath
 {
-  tls::vector<RatchetNode, 2> nodes;
+  std::vector<RatchetNode> nodes;
 
   TLS_SERIALIZABLE(nodes);
+  TLS_TRAITS(tls::vector_trait<2>{});
 };
 
 // struct {
@@ -55,7 +57,7 @@ struct KeyPackage
   HPKEPublicKey init_key;
   Credential credential;
   // TODO Extensions
-  tls::opaque<2> signature;
+  bytes signature;
 
   KeyPackage();
   KeyPackage(CipherSuite suite_in,
@@ -67,6 +69,7 @@ struct KeyPackage
   bool verify() const;
 
   TLS_SERIALIZABLE(version, cipher_suite, init_key, credential, signature);
+  TLS_TRAITS(tls::pass{}, tls::pass{}, tls::pass{}, tls::pass{}, tls::vector_trait<2>{});
 
   private:
   bytes to_be_signed() const;
@@ -87,18 +90,18 @@ struct KeyPackage
 //   opaque signature<0..255>;
 // } GroupInfo;
 struct GroupInfo {
-  tls::opaque<1> group_id;
+  bytes group_id;
   epoch_t epoch;
   RatchetTree tree;
-  tls::opaque<1> prior_confirmed_transcript_hash;
+  bytes prior_confirmed_transcript_hash;
 
-  tls::opaque<1> confirmed_transcript_hash;
-  tls::opaque<1> interim_transcript_hash;
+  bytes confirmed_transcript_hash;
+  bytes interim_transcript_hash;
   DirectPath path;
-  tls::opaque<1> confirmation;
+  bytes confirmation;
 
   LeafIndex signer_index;
-  tls::opaque<2> signature;
+  bytes signature;
 
   GroupInfo(CipherSuite suite);
   GroupInfo(bytes group_id_in,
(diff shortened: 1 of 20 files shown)


bifurcation:
@suhasHere - I think this is ready for review, modulo CI errors (which I think are just nits; tests pass locally for me).

include/messages.h
```diff
@@ -223,12 +231,17 @@ enum struct ContentType : uint8_t
...
+  TLS_SERIALIZABLE(content);
+  TLS_TRAITS(tls::variant<ProposalType>);
+};
 
-  static const ContentType type;
+struct ProposalID {
+  bytes id;
+  TLS_SERIALIZABLE(id);
+  TLS_TRAITS(tls::vector<1>);
 };
```
suhasHere, line 246: do we need both variations here .. Serializable and Traits ? wouldn't that end in double serialization of the id

suhasHere: or we need TLS_TRAITS{tls::pass)

include/messages.h
```diff
@@ -341,16 +364,23 @@ struct MLSPlaintext
...
+  bytes encrypted_sender_data;
+  bytes authenticated_data;
+  bytes ciphertext;
 
   TLS_SERIALIZABLE(group_id, epoch, content_type, sender_data_nonce,
                    encrypted_sender_data, authenticated_data, ciphertext);
+  TLS_TRAITS(tls::vector<1>,
+             tls::pass,
+             tls::pass,
+             tls::vector<1>,
```
suhasHere, line 380: don't we need tls::variant<ContentType> here (similar to line 353)

include/messages.h
```diff
@@ -329,6 +346,12 @@ struct MLSPlaintext
   bytes commit_auth_data() const;
 
   TLS_SERIALIZABLE(group_id, epoch, sender, authenticated_data, content, signature);
+  TLS_TRAITS(tls::vector<1>,
+             tls::pass,
+             tls::pass,
+             tls::vector<4>,
+             tls::variant<ContentType>,
+             tls::vector<2>)
```
suhasHere, line 354: should this be std::variant for ContentType ( IIUC we are getting rid of tls::variant)

src/messages.cpp
```diff
@@ -77,8 +104,13 @@ bytes
...
-  w << group_id << epoch << tree << confirmed_transcript_hash
-    << interim_transcript_hash << path << confirmation << signer_index;
+  tls::vector<1>{}.encode(w, group_id);
+  w << epoch << tree;
+  tls::vector<1>{}.encode(w, confirmed_transcript_hash);
+  tls::vector<1>{}.encode(w, interim_transcript_hash);
+  w << path;
+  tls::vector<1>{}.encode(w, confirmation);
+  w << signer_index;
   return w.bytes();
```
suhasHere, line 114: ah ! this side effect :-)

bifurcation:
Yeah, I wish it was smaller, but that's what we get for not catching this earlier :/

include/messages.h
bifurcation: No, because the content_type here is separated from the actual content. tls::variant is really intended for the case where you have the selector value followed immediately by the selected type.

src/messages.cpp
bifurcation: Yeah, it's ugly.  The trick here is that you need something ternary to combine (a) the stream (b) the trait and (c) the value.  You might be able to do this with `iomanip`-style [stream updaters](http://www.cplusplus.com/reference/iomanip/),  But then you need to (temporarily) persist the fact that there's a trait waiting to be processed, which could be non-trivial.  Or you could make temporary wrapper objects, but that's  challenging to do without causing additional allocations (requires the wrapper object to have a pointer or reference to the underlying data).  

```
// iomanip
w << path 
  << tls::vector<1>() << interim_transcript_hash
  << signer_index;

// wrapper
w << path << tls::vector<1>(confirmation) << signer_index;
```

After pondering a few of these options, my feeling was that having the traits as they are in the PR struck the right balance between simplicity and clarity in the default case (`TLS_TRAITS`).

include/messages.h
```diff
@@ -234,6 +240,8 @@ enum struct ContentType : uint8_t
 struct Proposal
 {
   std::variant<Add, Update, Remove> content;
+
+  static const ContentType type;
   TLS_SERIALIZABLE(content)
   TLS_TRAITS(tls::variant<ProposalType>)
```
suhasHere, line 246: where are we serializing ContentType?

bifurcation: That gets handled by the `tls::variant<ContentType>` inside `struct MLSPlaintext`

PR #95, opened 2020-09-01
Update the build system

bifurcation:
This PR is a pretty significant refactor of the tests and build system, based on lessons learned from the [SFrame implementation](https://github.com/cisco/sframe) setup.  There are several major changes here:

* Convert tests to doctest
* Remove the simulator app
* Simplify CMakeLists and Makefiles
* Enable clang-tidy more elegantly than before

Each of these should probably have been a PR of its own, but I was in a hurry.  When reviewing, it may be helpful to review individual commits rather than the whole diff.


bifurcation:
Ah, and I see that Travis CI is now failing because of the new doctest dependency.  Rather than fix that, I'm going to close this PR for now, and convert to GitHub actions first, since we have that figured out already.

PR #103, opened 2020-09-01
Add an HPKE library

bifurcation:
The idea here is to ultimately pull the underlying crypto dependencies out of the main MLS code base into a separable library, much like OpenSSL does.  About 80% of that is HPKE, with the remainder about 15% signature and 5% hashing.  Given that, together with the fact that HPKE is a separate spec that needs implementations, means that we'll just put the crypto stuff under the "hpke" heading for now.

So far, this just integrates the code.  Still need to move the relevant bits *out* of the main code and tie in the library instead.


lib/hpke/src/dhkem.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+std::unique_ptr<KEM::PublicKey>
+DHKEM::PrivateKey::public_key() const
+{
+  return group_priv->public_key();
+}
+
+///
+/// DHKEM::PrivateKey
+///
```
suhasHere, line 22: comment should be DHKEM ?

suhasHere (approved):
This is a massive PR . Overall nothing stood out as alarming. We should get a version of this ocne the test passes

include/mls/crypto.h
```diff
@@ -102,83 +79,65 @@ struct HPKEPublicKey
...
+  bytes data;
 
-  void set_cipher_suite(CipherSuite suite);
-  void set_signature_scheme(SignatureScheme scheme);
-  SignatureScheme signature_scheme() const;
-  bool verify(const bytes& message, const bytes& signature) const;
-  bytes to_bytes() const;
+  bool verify(const CipherSuite& suite,
+              const bytes& message,
+              const bytes& signature) const;
```
suhasHere, line 120: would be nice to have a version that takes in jus the scheme (and no suite) .. esp it helps for X509 work where i did add an API in that effect ..

cmd/test_gen/main.cpp
```diff
@@ -70,20 +74,20 @@ generate_crypto()
 
   // Construct a test case for each suite
   for (auto suite : suites) {
-    // HKDF-Extract
-    auto hkdf_extract_out =
-      hkdf_extract(suite, tv.hkdf_extract_salt, tv.hkdf_extract_ikm);
+    // kdf-Extract
+    auto kdf_extract_out =
+      suite.hpke->kdf->extract(tv.kdf_extract_salt, tv.kdf_extract_ikm);
```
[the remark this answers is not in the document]
bifurcation: I think it's OK to expose that MLS is using the KDF from HPKE.  The latest push refactors to use references to static objects instead of pointers, though.

include/mls/crypto.h
bifurcation: Nope.  SignatureScheme is no more!

bifurcation:
This latest push refactors from `std::unique_ptr<KEM>` (etc.) to `const KEM&`.  Using references allows polymorphism just like with pointers, allows the use of `.` syntax, and avoids a bunch of unnecessary duplicate objects being created.  The main risk in using reference members is that the object holding the reference will outlive the target of the reference, in which case undefined behavior ensues.  In this case, we avoid that risk by only making references to static objects, which by definition live forever.

suhasHere (approved):

PR #104, opened 2020-09-07
X509 support part 1 - ed25519

suhasHere:
This PR adds a starter X509 helper library with support for ED25519 signature and X509 Credential support as well. The X509 library is not complete , but can be considered as a base to add more things on .


include/mls/credential.h
```diff
@@ -2,6 +2,9 @@
 
 #include "mls/common.h"
 #include "mls/crypto.h"
+#include <memory>
+#include <openssl/x509.h>
+#include <openssl/x509v3.h>
```
bifurcation, line 7: Let's not use OpenSSL constructs in the public API for this.  I would like to at least maintain the pretense that we are independent of the underlying crypto library, even if we use OpenSSL internally, and callers are likely to use it to process what we provide.

include/mls/credential.h
```diff
@@ -38,6 +41,32 @@ struct BasicCredential
...
+
+// case x509:
+//     opaque cert_data<1..2^24-1>;
+struct X509Credential
+{
+
+  X509Credential() {}
+  explicit X509Credential(const std::vector<bytes>& chain_in);
+
+  std::vector<std::unique_ptr<X509Certificate>> chain;
```
bifurcation: Let's just make this `std::vector<bytes>`, and not even pretend to interpret the contents.

The Credential will need to interpret the leaf, to recognize the signature algorithm and extract the public key.  But it should just do that with OpenSSL on creation and store the public key thereafter.

include/mls/credential.h
```diff
@@ -38,6 +41,32 @@ struct BasicCredential
...
+  SignaturePublicKey public_key;
+  bytes identity;
```
bifurcation, line 59: I thought we agreed that we were going to suppress this field?

include/mls/x509_cert_helper.h
```diff
@@ -0,0 +1,31 @@
+#pragma once
```
bifurcation: This stuff can just go inside of `credential.cpp`.

include/mls/x509_cert_helper.h
```diff
@@ -0,0 +1,31 @@
...
+#include "bytes/bytes.h"
+#include <memory>
+
+using namespace std;
+using namespace bytes_ns;
+
+namespace mls {
+
+// TODO: move this inside X509Certificate to not mess up global space
+enum struct SignatureAlgorithm : uint16_t {
```
bifurcation: Nit: Suggest `X509SignatureAlgorithm`

bifurcation: Also, we should support ECDSA with P256/P384/P521 as well.

include/mls/x509_cert_helper.h
```diff
@@ -0,0 +1,31 @@
...
+	unknown = 0x0000,
+	ED25519 = 0x0001,
+	ED448 = 0x0002,
+};
+
+struct X509Certificate {
```
bifurcation: Not totally clear to me that you need a struct here, but maybe it's convenient.  Seems like it should just be:

```
struct Certificate {
  explicit Certificate(const bytes& der);
  static std::vector<Certificate> parse_chain(const bytes& der_chain);

  const bytes& raw() const;
  const SignaturePublicKey& public_key() const;
  bool signed_by(const Certificate& parent) const;

  private:
  typed_unique_ptr<X509> _x509;
  bytes _raw;
  SignaturePublicKey _public_key;

  Certificate(X509* x509_in, bytes raw_in);
};
```

... since that's all we need inside of MLS.

src/x509_cert_helper.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+static std::runtime_error
+openssl_error()
+{
+	uint64_t code = ERR_get_error();
+	return std::runtime_error(ERR_error_string(code, nullptr));
+}
+
+
+struct OpenSSLCertificate : public X509Certificate {
```
bifurcation: I wouldn't bother with the layer of indirection.  Just have one real certificate class; I'm agnostic as to whether you call it `OpenSSLCertificate` or just `Certificate`.

src/x509_cert_helper.cpp
```diff
@@ -0,0 +1,123 @@
...
+	explicit OpenSSLCertificate(bytes cert_in) {
+		cert_raw = std::move(cert_in);
+		const unsigned char *buf = cert_raw.data();
+		X509 *tmp = d2i_X509(nullptr, &buf, cert_raw.size());
```
bifurcation: 1. Always use `auto` for declarations.  Here and on the previous line `const auto*`
2. You should immediately wrap this in a smart pointer, `make_typed_unique(...)`.  If you need to pass it, you can then use `.release()`.

However, I notice that `typed_unique_ptr` is up in the `hpke` library now.  Which might point to a need to move this code up there as well.

src/x509_cert_helper.cpp
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	bytes subject_name() const override
+	{
+		return bytes();
+	}
+
+private:
+	bytes cert_raw;
+	X509* certificate_ptr;
```
bifurcation: This *definitely* needs to be a smart pointer, probably `typed_unique_ptr<X509>`.

include/mls/credential.h
```diff
@@ -38,6 +42,25 @@ struct BasicCredential
...
+///
+
+// case x509:
+//     opaque cert_data<1..2^24-1>;
+struct X509Credential
+{
+  X509Credential() {}
+  explicit X509Credential(const std::vector<bytes>& raw_chain);
+
+  std::vector<std::shared_ptr<Certificate>> chain;
```
bifurcation, line 56: This should just be `std::vector<Certificate>`

include/mls/credential.h
```diff
@@ -38,6 +42,25 @@ struct BasicCredential
...
+///
+/// X509 Credential
+///
+
+// case x509:
+//     opaque cert_data<1..2^24-1>;
+struct X509Credential
+{
+  X509Credential() {}
+  explicit X509Credential(const std::vector<bytes>& raw_chain);
```
bifurcation, line 54: Nit: Instead of `raw`, use `der` to specify the format.

include/mls/credential.h
```diff
@@ -38,6 +42,25 @@ struct BasicCredential
...
+
+  std::vector<std::shared_ptr<Certificate>> chain;
+  SignaturePublicKey public_key;
+
+  TLS_SERIALIZABLE(public_key)
```
bifurcation, line 59: This is wrong.  You need to serialize the whole chain.

lib/CMakeLists.txt
```diff
@@ -1,3 +1,3 @@
 add_subdirectory(bytes)
 add_subdirectory(hpke)
-add_subdirectory(tls_syntax)
+add_subdirectory(tls_syntax)
```
bifurcation, line 3: Please revert this change.  I think you just need a line break at the end.

lib/hpke/include/hpke/certificate.h
```diff
@@ -0,0 +1,50 @@
+#pragma once
+
+#include "bytes/bytes.h"
```
bifurcation, line 3:
```suggestion
#include <bytes/bytes.h>
```

lib/hpke/include/hpke/certificate.h
```diff
@@ -0,0 +1,50 @@
...
+#include <memory>
+#include <optional>
+
+using namespace bytes_ns;
+using namespace std;
```
bifurcation, line 8: Never do `using namespace std;`.  In general `using namespace` is to be avoided; we just do it for `bytes` for convenience.

lib/hpke/include/hpke/certificate.h
```diff
@@ -0,0 +1,50 @@
...
+
+namespace hpke {
+
+// Wrapper around X509 OpenSsl API
```
bifurcation, line 12:
```suggestion
// Wrapper around X509 OpenSSL API
```

lib/hpke/include/hpke/certificate.h
```diff
@@ -0,0 +1,50 @@
...
+      ED448 = 0x0002,
+    };
+
+    struct PublicKey
+    {
+      bytes data;
+    };
+
+    PublicKey pkey;
+    ID algorithm;
```
bifurcation, line 31:
```suggestion
    const ID public_key_algorithm;
```

... to distinguish it from the algorithm used to sign the cert.

lib/hpke/include/hpke/certificate.h
```diff
@@ -0,0 +1,50 @@
...
+  };
+
+  Certificate();
```
bifurcation, line 34: Is this meaningful?  Necessary to make `unique_ptr` work?

lib/hpke/include/hpke/certificate.h
```diff
@@ -0,0 +1,50 @@
...
+  explicit Certificate(const bytes& der);
+  ~Certificate();
+
+  Signature::ID signature_algorithm() const;
+
+  Signature::PublicKey public_key() const;
```
bifurcation, line 40: These two methods are unnecessary if you have the public fields above.

lib/hpke/src/certificate.cpp
```diff
@@ -0,0 +1,113 @@
...
+#include "openssl_common.h"
+#include <openssl/err.h>
+#include <openssl/x509.h>
+
+namespace hpke {
+
+///
+/// Private Implementation
+///
+///
```
bifurcation, line 11:
```suggestion
```

lib/hpke/src/certificate.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+struct Certificate::Internals
+{
+
+  static X509* der_to_x509(const bytes& der);
```
bifurcation, line 16: You can just define the methods inline here (within a .cpp).

bifurcation:
```suggestion
  static X509* parse(const bytes& der);
```

lib/hpke/src/certificate.cpp
```diff
@@ -0,0 +1,113 @@
...
+      return Signature::ID::ED25519;
+    case EVP_PKEY_ED448:
+      return Signature::ID::ED448;
+    default:
+      // TODO (Suhas): Add support for ECDSA curves
+      return Signature::ID::unknown;
+  }
+}
+
+std::optional<bytes>
```
bifurcation, line 57: Just make this `bytes` and throw on failure.

lib/hpke/src/certificate.cpp
```diff
@@ -0,0 +1,113 @@
...
+  bytes public_key;
+  auto algo = signature_algorithm();
+  switch (algo) {
+    case Signature::ID::ED448:
+    case Signature::ID::ED25519: {
+      auto key =
+        make_typed_unique<EVP_PKEY>(X509_get_pubkey(openssl_cert.get()));
+      size_t raw_len = 0;
+      if (1 != EVP_PKEY_get_raw_public_key(key.get(), nullptr, &raw_len)) {
+        break;
```
bifurcation, line 69:
```suggestion
        throw openssl_error();
```

lib/hpke/src/certificate.cpp
```diff
@@ -0,0 +1,113 @@
...
+      }
+      public_key.resize(raw_len);
+      auto* data_ptr = public_key.data();
+      if (1 != EVP_PKEY_get_raw_public_key(key.get(), data_ptr, &raw_len)) {
+        break;
```
bifurcation, line 74:
```suggestion
        throw openssl_error();
```

lib/hpke/include/hpke/certificate.h
```diff
@@ -0,0 +1,50 @@
...
+using namespace bytes_ns;
+using namespace std;
+
+namespace hpke {
+
+// Wrapper around X509 OpenSsl API
+struct Certificate
+{
+
+  struct Signature
```
bifurcation, line 16: Reuse `hpke::Signature`.

suhasHere:
Closing since it will be replaced by #106

PR #106, opened 2020-09-19
X509 - support eddsa sigs

suhasHere:


src/credential.cpp
```diff
@@ -49,4 +69,16 @@ Credential::basic(const bytes& identity, const SignaturePublicKey& public_key)
   return cred;
 }
 
+Credential
+Credential::x509(const std::vector<bytes>& der_chain)
+{
+  if (der_chain.empty()) {
+    throw std::invalid_argument("empty cert chain");
+  }
+
```
suhasHere, line 78: this check is superflous .. will delete it in next commit

include/mls/credential.h
```diff
@@ -38,6 +39,20 @@ struct BasicCredential
...
+struct X509Credential
+{
+  X509Credential() {}
+
+  explicit X509Credential(const std::vector<bytes>& der_chain);
+
+  SignaturePublicKey public_key;
+  std::vector<hpke::Certificate> chain;
+
+  TLS_SERIALIZABLE(public_key)
```
bifurcation, line 51: This is wrong.  You need to write a custom serializer that concatenates the chain and then marshals it as a `tls::vector<2>`.  And the reverse in the deserialize direction.

include/mls/credential.h
```diff
@@ -38,6 +39,20 @@ struct BasicCredential
   static const CredentialType type;
 };
 
+struct X509Credential
+{
+  X509Credential() {}
+
+  explicit X509Credential(const std::vector<bytes>& der_chain);
```
bifurcation, line 46: You can offer this if you want, but what you really want (especially on deserialize) is `X509Credential(const bytes& der_chain)`.  Unless we're going to change the spec to have a separate `opaque` per cert.  Which might be a good idea.

test/credential.cpp
```diff
@@ -15,3 +16,38 @@ TEST_CASE("Basic Credential")
   REQUIRE(cred.identity() == user_id);
   REQUIRE(cred.public_key() == pub);
 }
+
+std::vector<std::string>
+read_hex_file(const std::string& filename)
```
bifurcation, line 21: Again, don't use files.

lib/hpke/include/hpke/certificate.h
```diff
@@ -0,0 +1,32 @@
...
+  explicit Certificate(const bytes& der);
+  Certificate(const Certificate& other);
+  Certificate& operator=(const Certificate& other);
+  Certificate(Certificate&& other) noexcept;
+  ~Certificate();
+
+  // bool valid_from(const Certificate& parent);
+
+  const Signature::ID public_key_algorithm;
+  const bytes public_key;
```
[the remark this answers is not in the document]
suhasHere: as agreed, i will be bring back the X509Signature struct instead of depending on hpke:: variants for this puporses.

include/mls/credential.h
```diff
@@ -38,6 +39,20 @@ struct BasicCredential
...
+
+  SignaturePublicKey public_key;
+  std::vector<hpke::Certificate> chain;
```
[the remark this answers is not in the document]
suhasHere: i am inclined to leaving this as it is for now as well.  but i wonder we need to convert form bytes to a concrete object to extract publicKey and so on .. One way to do is the vector<hpke::Certificate> , but i am not clear how the sme can be achieved with just vector<bytes>

What am i missing here ?

include/mls/credential.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "hpke/certificate.h"
 #include "mls/common.h"
```
suhasHere, line 4: this needs to go away. Fixed locally.

test/credential.cpp
```diff
@@ -15,3 +16,85 @@ TEST_CASE("Basic Credential")
...
+  std::vector<bytes> der_in(1);
+  der_in[0] = leaf_der;
+
+  auto actual = Credential::x509(der_in);
+  CHECK(actual.public_key().data.size() != 0);
+
+  auto marshalled = tls::marshal(actual);
+  Credential expected;
+  tls::unmarshal(marshalled, expected);
+  CHECK(actual.public_key() == expected.public_key());
```
bifurcation, line 99:
```suggestion
  auto unmarshaled = tls::get<Credential>(marshalled);
  CHECK(original == unmarshaled);
```

test/credential.cpp
```diff
@@ -15,3 +16,85 @@ TEST_CASE("Basic Credential")
...
+             "36333230345a3000302a300506032b65700321009d28cb0fdb2da572066ba1c90"
+             "dedbf8010ba85fba4bffa8fffe67e26d5515252a320301e300e0603551d0f0101"
+             "ff0404030202a4300c0603551d130101ff04023000300506032b6570034100b59"
+             "e47d1f9de4d595ca2d0a2a1dbc6a59eacfa3e5958d315d3809d6f7abba0474643"
+             "a620e81b91ec987b30214a30818f97e47f21ad15d3b304a6ebf142b36105");
+  std::vector<bytes> der_in(1);
+  der_in[0] = leaf_der;
+
+  auto actual = Credential::x509(der_in);
+  CHECK(actual.public_key().data.size() != 0);
```
bifurcation, line 94:
```suggestion
  auto original = Credential::x509(der_in);
  CHECK(original.public_key().data.size() != 0);
```

test/credential.cpp
```diff
@@ -15,3 +16,85 @@ TEST_CASE("Basic Credential")
...
+  const auto leaf_der =
+    from_hex("3081de308191a00302010202110089f2d252744135d5f3f005f1edc609d230050"
+             "6032b65703000301e170d3230303932313036333230345a170d32303039323230"
+             "36333230345a3000302a300506032b65700321009d28cb0fdb2da572066ba1c90"
+             "dedbf8010ba85fba4bffa8fffe67e26d5515252a320301e300e0603551d0f0101"
+             "ff0404030202a4300c0603551d130101ff04023000300506032b6570034100b59"
+             "e47d1f9de4d595ca2d0a2a1dbc6a59eacfa3e5958d315d3809d6f7abba0474643"
+             "a620e81b91ec987b30214a30818f97e47f21ad15d3b304a6ebf142b36105");
+  std::vector<bytes> der_in(1);
+  der_in[0] = leaf_der;
```
bifurcation, line 91:
```suggestion
  auto der_in = std::vector<bytes>{leaf_der};
```

suhasHere: ah ! how did i miss this .. good point, will updatte

lib/hpke/include/hpke/certificate.h
```diff
@@ -0,0 +1,50 @@
...
+
+struct Certificate
+{
+
+private:
+  struct ParsedCertificate;
+  std::unique_ptr<ParsedCertificate> parsed_cert;
+
+public:
+  struct X509Signature
```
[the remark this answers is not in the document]
suhasHere: @bifurcation  This goes back to the conversation on converting hpke::signature to SignaturePublicKey type without having access to suite for this . We concluded its fine to define one specific for X509.

include/mls/credential.h
```diff
@@ -38,6 +38,29 @@ struct BasicCredential
...
+
+  X509Credential() {}
+
+  explicit X509Credential(std::vector<CertData> der_chain_in);
+
+  std::vector<CertData> der_chain;
+  SignaturePublicKey public_key;
+
+  TLS_SERIALIZABLE(der_chain, public_key)
+  TLS_TRAITS(tls::vector<4>, tls::pass)
```
bifurcation, line 59:
```suggestion
  TLS_SERIALIZABLE(der_chain)
  TLS_TRAITS(tls::vector<4>)
```

lib/hpke/include/hpke/certificate.h
```diff
@@ -0,0 +1,31 @@
...
+
+#include <memory>
+
+using namespace bytes_ns;
+
+namespace hpke {
+
+struct Certificate
+{
+private:
```
bifurcation, line 14: Nit: Move this to the bottom of the struct, and then you won't need the `public:` annotation.

include/mls/credential.h
```diff
@@ -38,6 +38,29 @@ struct BasicCredential
...
+    TLS_SERIALIZABLE(data);
+    TLS_TRAITS(tls::vector<2>)
+  };
+
+  X509Credential() {}
+
+  explicit X509Credential(std::vector<CertData> der_chain_in);
+
+  std::vector<CertData> der_chain;
+  SignaturePublicKey public_key;
```
bifurcation, line 56: 1. This should either be const or private and exposed via a method.
2. We should expose the signature algorithm here and verify it in `KeyPackage::verify()`.  Please file an issue to do this in a follow-on.

bifurcation (commented):
Couple of last minor things.

lib/hpke/include/hpke/certificate.h
suhasHere: leaving it as is to have compiler construct things in the order we need here.

include/mls/credential.h
```diff
@@ -38,6 +38,37 @@ struct BasicCredential
...
+{
+  struct CertData
+  {
+    bytes data;
+
+    TLS_SERIALIZABLE(data);
+    TLS_TRAITS(tls::vector<2>)
+  };
+
+  X509Credential() {}
```
bifurcation, line 51:
```suggestion
  X509Credential() = default;
```

include/mls/credential.h
```diff
@@ -38,6 +38,37 @@ struct BasicCredential
...
+
```
bifurcation, line 52:
```suggestion
```

include/mls/credential.h
```diff
@@ -38,6 +38,37 @@ struct BasicCredential
...
+  explicit X509Credential(std::vector<CertData> der_chain_in);
+
+  SignaturePublicKey public_key() const;
+
+  std::vector<CertData> der_chain;
+
```
bifurcation, line 58:
```suggestion
```

include/mls/credential.h
```diff
@@ -38,6 +38,37 @@ struct BasicCredential
...
+  static const CredentialType type;
+
+private:
+  SignaturePublicKey _public_key;
+};
+
+// custom marshal/unmarshal operations
```
bifurcation, line 65:
```suggestion
```

lib/hpke/src/certificate.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include <hpke/signature.h>
+#include <openssl/x509.h>
+
+#include "group.h"
+#include "openssl_common.h"
+
+namespace hpke {
+
+///
+/// Certificate X509 Impl
```
bifurcation, line 11:
```suggestion
/// ParsedCertificate
```

PR #107, opened 2020-09-21
Streamline and complete the session API

bifurcation:
The Session API has been kind of pro-forma for a while.  This PR makes several changes that make it more usable.

* Generate secrets internally to Session, rather than making the caller provide entropy
* Don’t commit on every operation, so that callers can coalesce operations into one Commit
* Explicit join process (Client -> PendingJoin -> Session)
* Hide the internals of API classes (as a prelude to slimming down the set of public headers)
* Implement and expose the MLS exporter

There is still some work to do to make this fully functional (e.g., exposing the roster) and fully minimal (e.g., making headers private).  But this seemed like a good start.

### DIFF
--- a/cmd/api_example/main.cpp
+++ b/cmd/api_example/main.cpp
@@ -66,10 +66,7 @@ main() // NOLINT(bugprone-exception-escape)
   // Alice adds Bob and Charlie to the session
   auto add_bob = alice_session.add(bob_join.key_package());
   auto add_charlie = alice_session.add(charlie_join.key_package());
-
-  alice_session.handle(add_bob);
-  alice_session.handle(add_charlie);
-  auto [welcome, commit] = alice_session.commit();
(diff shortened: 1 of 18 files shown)


cmd/api_example/main.cpp
```diff
@@ -70,74 +43,69 @@ verify(const std::string& label, Session& alice, Session& bob)
...
 
-  // Charlie posts a KeyPackage
-  auto kpC1 = charlie.fresh_key_package();
-
-  // Alice adds Charlie to the session
-  bytes add;
-  std::tie(welcome, add) = sessionA.add(random_bytes(32), kpC1);
+  // Bob and Charlie post KeyPackages
+  auto bob_join = bob_client.start_join();
+  auto charlie_join = charlie_client.start_join();
```
suhasHere, line 65: not that it is needed , but as much as i think auto is a good thing, knowing the type here will help readability without needing to look at the source code .

suhasHere (approved):

PR #109, opened 2020-09-22
build 1.1 openssl

suhasHere:


.github/workflows/build.yml
```diff
@@ -23,11 +23,20 @@ jobs:
...
+        rm -rf /usr/local/etc/openssl@1.1
         brew install pkgconfig doctest
-        brew reinstall openssl
+        brew reinstall openssl@1.1
+  
+    - name: dependencies (ubuntu)
+      if: ${{ matrix.os == 'ubuntu-latest' }}
+      run: |
+        brew install pkgconfig doctest
+        brew reinstall openssl@1.1
```
bifurcation, line 39: You might not need this line.

cmd/api_example/main.cpp
```diff
@@ -139,6 +139,6 @@ main() // NOLINT(bugprone-exception-escape)
 
   verify("remove A->C", sessionA, sessionC);
 
-  std::cout << "ok" << std::endl;
+  std::cout << "ok, all good" << std::endl;
```
bifurcation, line 142: I don't think this change is needed :)

.github/workflows/build.yml
```diff
@@ -23,11 +23,20 @@ jobs:
...
-      if: ${{ matrix.os == 'macos-latest' || matrix.os == 'ubuntu-latest' }}
+    - name: dependencies (macos)
+      if: ${{ matrix.os == 'macos-latest' }}
       run: |
+        brew uninstall openssl
+        rm -rf /usr/local/etc/openssl
+        rm -rf /usr/local/etc/openssl@1.1
         brew install pkgconfig doctest
-        brew reinstall openssl
+        brew reinstall openssl@1.1
```
bifurcation, line 33: Since we just uninstalled it, could this just go up on the `install` line above?

.github/workflows/build.yml
```diff
@@ -23,11 +23,20 @@ jobs:
...
     - uses: actions/checkout@v1
     
-    - name: dependencies (ubuntu and macos)
-      if: ${{ matrix.os == 'macos-latest' || matrix.os == 'ubuntu-latest' }}
+    - name: dependencies (macos)
+      if: ${{ matrix.os == 'macos-latest' }}
       run: |
+        brew uninstall openssl
+        rm -rf /usr/local/etc/openssl
+        rm -rf /usr/local/etc/openssl@1.1
```
bifurcation, line 31: Are these lines really needed?

cmd/api_example/main.cpp
suhasHere: haha .. good point, will remove

.github/workflows/build.yml
suhasHere: i added to ensure it goes clean .. but may be its fine and they get overwritten

bifurcation (approved):

PR #111, opened 2020-09-25
Identify ciphers and expose credentials

bifurcation:
Two major changes here:

* Crypto objects from the `hpke` library now all have a `const ID id` field 
* Credential objects now expose the underlying typed credential objects

### DIFF
--- a/lib/hpke/include/hpke/signature.h
+++ b/lib/hpke/include/hpke/signature.h
@@ -21,7 +21,6 @@ struct Signature
   template<Signature::ID id>
   static const Signature& get();
 
-  Signature(ID id_in);
   virtual ~Signature() = default;
 
   struct PublicKey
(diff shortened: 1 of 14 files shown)


lib/hpke/include/hpke/signature.h
```diff
@@ -21,6 +21,7 @@ struct Signature
   template<Signature::ID id>
   static const Signature& get();
 
+  Signature(ID id_in);
   virtual ~Signature() = default;
```
suhasHere, line 25: This allows Signature to be created in 2 different ways. One via this constructor and one a concrete version via get<T>() API.

bifurcation: Good catch, moved those ctors to be protected, since they should only be called by subclasses.

suhasHere (approved):

PR #112, opened 2020-09-26
Support Roster API on Session

suhasHere:


src/state.cpp
```diff
@@ -580,6 +580,21 @@ State::do_export(const std::string& label,
   return _suite.expand_with_label(secret, "exporter", context, size);
 }
 
+std::vector<Credential>
+State::get_leaf_credentials() const
+{
+  std::vector<Credential> creds;
```
bifurcation, line 586: You might get a little efficiency by allocating this to full size (`_tree.size()`) and then truncating it at the end (`creds.resize(leaf_count)`, where `leaf_count` is the number of non-blank leaves you found)

src/state.cpp
```diff
@@ -580,6 +580,21 @@ State::do_export(const std::string& label,
...
+
+  for (uint32_t i = 0; i < _tree.size().val; i++) {
+    const auto& ln = _tree.node_at(LeafIndex{ i });
```
bifurcation, line 589: This is what `TreeKEMPublicKey::key_package(LeafIndex index)` is for.  `node_at` should probably be private.

src/state.cpp
```diff
@@ -580,6 +580,21 @@ State::do_export(const std::string& label,
...
+    if (ln.node.has_value()) {
```
bifurcation, line 590: I would reverse this logic here:

```
if (!ln.node.has_value) {
  continue;
}

// push_back
```

bifurcation (commented):
In addition to these comments, there's another change that should get bundled in here.  Namely, the index provided to `Session::remove()` should refer to the index **in the roster**, not among the leaves of the tree.  This is because the caller only has access to the roster, not the tree.  So `Session::remove()` will need to map that index back to a tree index before calling `State::remove()`, e.g., by counting up from the left.

src/state.cpp
```diff
@@ -153,7 +153,19 @@ State::update(const bytes& leaf_secret)
...
+    const auto& kp = _tree.key_package(LeafIndex{ i });
+    if (kp.has_value()) {
+      if (non_blank_leaves == removed.val) {
+        return sign({ Remove{ LeafIndex{ i } } });
+      }
+      ++non_blank_leaves;
+    }
+  }
+
+  throw InvalidParameterError("Leaf Index mismatch");
```
bifurcation, line 168: Sorry if I was unclear before.  I think you have the algorithm right now, but I would still prefer things split out.  Something like:

```
struct State {
  struct RosterIndex : public Uint32{};
  // ^^^ suggest putting this here because it's not meaningful without the tree.
  // But you could also make it a stand-alone type, parallel to LeafIndex and NodeIndex.
  // In any case, we should use the type system to avoid confusion here.

  LeafIndex leaf_for_roster_entry(RosterIndex index); // <-- this algorithm
  MLSPlaintext remove(RosterIndex removed) const {
    return remove(leaf_for_roster_entry(removed));
  }
}
```

src/state.cpp
```diff
@@ -580,6 +592,25 @@ State::do_export(const std::string& label,
...
+{
+  std::vector<Credential> creds(_tree.size().val);
+  uint32_t leaf_count = 0;
+
+  for (uint32_t i = 0; i < _tree.size().val; i++) {
+    const auto& kp = _tree.key_package(LeafIndex{ i });
+    if (!kp.has_value()) {
+      continue;
+    }
+    creds[i] = kp->credential;
```
bifurcation, line 606: This should assign to `cred[leaf_count]`.

bifurcation (commented):
It would be good to have a more explicit test, something like:

1. Create a 5-member group {A, B, C, D, E}
2. Remove B and E by RosterIndex (1 and 2, respectively)
3. Verify that the resulting roster is {A, C, E}

src/state.cpp
suhasHere: good point .. will do

test/state.cpp
```diff
@@ -154,6 +160,64 @@ TEST_CASE_FIXTURE(StateTest, "Full Size Group")
...
+  // remove member at position 2
+  auto remove_2 = states[0].remove(RosterIndex{ 2 });
+  states[0].handle(remove_2);
+
+  // commit to new state
+  auto [commit, welcome, new_state] = states[0].commit(fresh_secret());
+  silence_unused(welcome);
+
+  auto roster_ref = new_state.roster();
+  REQUIRE(3 == roster_ref.size());
```
bifurcation, line 210: You can do better than this by checking the actual contents of the roster.
```
  auto expected_roster = std::vector<Credential>{
    key_packages[0].credential,
    key_packages[2].credential,
    key_packages[4].credential,
  };
  REQUIRE(expected_roster = new_state.roster());
```

I think you will then find the bug I mention above :)

src/state.cpp
```diff
@@ -150,6 +150,27 @@ State::update(const bytes& leaf_secret)
...
 
+LeafIndex
+State::leaf_for_roster_entry(RosterIndex index) const
+{
+  auto non_blank_leaves = uint32_t (0);
+
+  for (auto i = LeafIndex{0}; i < _tree.size(); i.val++) {
+    const auto& kp = _tree.key_package(i);
+    if (!kp.has_value()) { continue; }
+    if (non_blank_leaves == index.val) { return i; }
```
bifurcation, line 161:
```suggestion
    if (!kp.has_value()) {
      continue;
    }
    if (non_blank_leaves == index.val) {
      return i;
    }
```

test/state.cpp
```diff
@@ -154,6 +160,84 @@ TEST_CASE_FIXTURE(StateTest, "Full Size Group")
   }
 }
 
+TEST_CASE_FIXTURE(StateTest, "Roster Updates")
```
bifurcation, line 163: Still should use RunninGroupTest

bifurcation (commented):
Almost there

bifurcation (approved):

PR #113, opened 2020-09-26
verify signature id in KeyPackage for X509 Credentials

suhasHere:
Issue #110

### DIFF
--- a/include/mls/credential.h
+++ b/include/mls/credential.h
@@ -60,6 +60,9 @@ struct X509Credential
 
 private:
   SignaturePublicKey _public_key;
+  hpke::Signature::ID _signature_algorithm;
+
+  friend struct KeyPackage;
 };
 
 tls::ostream&
--- a/src/core_types.cpp
+++ b/src/core_types.cpp
@@ -104,6 +104,14 @@ KeyPackage::verify() const
 {
   auto tbs = to_be_signed();
   auto identity_key = credential.public_key();
+
+  if (CredentialType::x509 == credential.type()) {
+    const auto& cred = credential.get<X509Credential>();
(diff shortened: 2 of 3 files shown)


include/mls/credential.h
```diff
@@ -60,6 +60,9 @@ struct X509Credential
 
 private:
   SignaturePublicKey _public_key;
+  hpke::Signature::ID _signature_algorithm;
```
bifurcation, line 63: This exposes the HPKE library, which I would rather not do.  But on the other hand, hiding it involves some contortions.  Not sure what the best plan is here.

include/mls/crypto.h
```diff
@@ -10,6 +10,15 @@
 #include <vector>
 
 namespace mls {
+/// Signature Code points
```
bifurcation, line 13:
```suggestion
// Signature Code points, borrowed from RFC 8446
```

src/core_types.cpp
```diff
@@ -2,6 +2,24 @@
 
 namespace mls {
 
+hpke::Signature::ID
+hpke_signature_algorithm(SignatureScheme scheme)
```
bifurcation, line 6: You should only need one direction of this conversion, `hpke -> tls`.  Just delete this method and in the check below:
```
BEFORE: to_hpke(cred._signature_scheme) != cipher_suite.get().sig.id
AFTER : cred._signature_scheme != to_tls(cipher_suite.get().sig.id)
```

I would also put that conversion in `crypto.h` / `crypto.cpp`.  That will expose `hpke::Signature::ID` in the short run, but (a) you should be able to forward-declare it, and (b) we can make it internal in the longer run.

bifurcation (approved):

PR #114, opened 2020-10-01
C api wrapper

TobiasWen:
Greetings,

during the course of my master's thesis, which is dealing with the suitability of the current state of mls in distributed peer to peer environments, I needed to develop a wrapper for the mlspp api for linkage against a C library. In doing so I don't want to withhold the result and maybe you have a use for it or simply want to store it in another branch for possible future use. Though my time is limited I will definitely respond to any change suggestions during review.


bifurcation:
Thanks for posting this, @TobiasWen!  I think we might have some API evolution as the spec finalizes and we upgrade mlspp to match, so I might be inclined to hold off on merging this for a bit.  But it will probably be helpful to have a binding like this in the long run.

bifurcation:
Closing this for now.  As we get the protocol stabilized and MLSpp upgraded, it might be worth resurrecting later.

PR #118, opened 2020-10-10
External commit

bifurcation:
This PR implements an elaborated form of https://github.com/mlswg/mls-protocol/pull/406/.  It seems to work in that the following test passes:

```C++
TEST_CASE_FIXTURE(StateTest, "External Join")
{
  // Initialize the creator's state
  auto first0 =
    State{ group_id, suite, init_privs[0], identity_privs[0], key_packages[0] };
  auto group_key_package = first0.group_key_package();

  // Initialize the second participant as an external joiner
  auto [commit, second0] = State::external_join(
    fresh_secret(), identity_privs[1], key_packages[1], group_key_package);

  // Creator processes the commit
  auto first1 = first0.handle(commit).value();

  auto group = std::vector<State>{ first1, second0 };
  verify_group_functionality(group);
}
```

This work validates that external commits are implementable, and highlights a few things that need to be added to the PR to make it complete:
* Defining a new SenderType value for this purpose
* Defining the ExternalInit proposal type and its handling
* Defining the minimum requirements for an external commit:
    * Must have an Add proposal that adds the new member
    * Must have an ExternalInit proposal
    * Must have a path
* Defining how the signature on an external commit is validated
    * Here, we use the public key in Commit.path.leaf_key_package.credential, which ensures Commit.path is populated
    * You could also have it be a reference to the Add where the joiner is added, which would ensure that there's an Add proposal for the joiner

### DIFF
--- a/include/mls/core_types.h
+++ b/include/mls/core_types.h
@@ -70,6 +70,7 @@ struct ExtensionList
   }
 
   bool has(uint16_t type) const;
+  ExtensionList for_group() const;
 
   TLS_SERIALIZABLE(extensions)
   TLS_TRAITS(tls::vector<2>)
--- a/include/mls/crypto.h
+++ b/include/mls/crypto.h
@@ -107,6 +107,10 @@ struct HPKEPublicKey
                          const bytes& aad,
                          const bytes& pt) const;
 
+  std::tuple<bytes, bytes> do_export(CipherSuite suite,
+                                     const std::string& label,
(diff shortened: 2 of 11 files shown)


bifurcation:
The spec now covers the questions posed in the first comment on this PR.  I have updated it to match the spec and rebased it on the current `main` branch, so it should be ready for review.

suhasHere (approved):

PR #135, opened 2020-11-02
Add membership_tag and confirmation_tag

### LINKED ISSUE #127: Authenticate membership on MLSPlaintext
Recent spec versions have added a `membership_tag` to MLSPlaintext that authenticates the sender's membership in the group before any processing is done on the MLSPlaintext.

https://github.com/mlswg/mls-protocol/blob/master/draft-ietf-mls-protocol.md#content-authentication

### LINKED ISSUE #128: Move Commit signature under confirmation_tag
The confirmation tag is used by endpoints to verify that they have arrived at the same key schedule.  Recent spec versions have moved to include the signature under the confirmation tag.  This will just entail adapting the signing/verification logic so that the signature is included in the confirmed transcript hash.

This has already been prototyped in #118, specifically:
https://github.com/cisco/mlspp/pull/118/commits/d37b5370321dee9c2eb3bae448d3283a73a2a13c

bifurcation:
Fixes #127 
Fixes #128

Depends on #134

### DIFF
--- a/src/state.cpp
+++ b/src/state.cpp
@@ -247,7 +247,7 @@ State::commit(const bytes& leaf_secret) const
     next._confirmed_transcript_hash,
     next._interim_transcript_hash,
     next._extensions,
-    std::get<CommitData>(pt.content).confirmation,
+    pt.confirmation_tag.value().mac_value,
   };
   group_info.sign(_index, _identity_priv);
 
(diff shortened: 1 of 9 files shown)


src/state.cpp
```diff
@@ -247,7 +248,7 @@ State::commit(const bytes& leaf_secret) const
     next._confirmed_transcript_hash,
     next._interim_transcript_hash,
     next._extensions,
-    std::get<CommitData>(pt.content).confirmation,
+    pt.confirmation_tag.value().mac_value,
```
suhasHere, line 251: just to double check, are we sure that we will always have confirmation_tag here given it is an optional value ?

suhasHere (commented):
one clarification question .. Otherwise LGTM

suhasHere (approved):

src/state.cpp
bifurcation: Yep.  It gets set in `ratchet_and_sign` above.

PR #138, opened 2020-11-04
Library pkgconfig generation

palerikm:
Added:
- Install target (make install, from your build directory)
- Generate a pkgconfig file to make it easier to link against this library
- Simple semantic versioning and git sha hash to identify build

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,8 +1,27 @@
 cmake_minimum_required(VERSION 3.12)
+set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake ${CMAKE_MODULE_PATH})
+
+
+###
+###  Semantic versioning
+###
+include ( GetGitRevisionDescription )
+git_describe(VERSION --tags)
+get_git_head_revision(REFSPEC HASHVAR)
+
+string(REGEX REPLACE "^v([0-9]+)\\..*" "\\1" MLSPP_VERSION_MAJOR "${VERSION}")
+string(REGEX REPLACE "^v[0-9]+\\.([0-9]+).*" "\\1" MLSPP_VERSION_MINOR "${VERSION}")
+string(REGEX REPLACE "^v[0-9]+\\.[0-9]+\\.([0-9]+).*" "\\1" MLSPP_VERSION_PATCH "${VERSION}")
+string(REGEX REPLACE "^v[0-9]+\\.[0-9]+\\.[0-9]+(.*)" "\\1" MLSPP_VERSION_SHA1 "${VERSION}")
+mark_as_advanced(MLSPP_VERSION_MAJOR MLSPP_VERSION_MINOR MLSPP_VERSION_PATCH)
+set ( MLSPP_VERSION "${MLSPP_VERSION_MAJOR}.${MLSPP_VERSION_MINOR}.${MLSPP_VERSION_PATCH}" )
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/cmake/version.c.in
+        ${CMAKE_CURRENT_BINARY_DIR}/version.cpp)
+set(version_file "${CMAKE_CURRENT_BINARY_DIR}/version.cpp")
 
 project(mlspp
-  VERSION 0.1
-  LANGUAGES CXX
+        VERSION "${MLSPP_VERSION}"
+        LANGUAGES CXX
 )
 
 option(CLANG_TIDY "Perform linting with clang-tidy" OFF)
@@ -52,6 +71,16 @@ endif()
 add_subdirectory(lib)
 
 # External libraries
+if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+  find_package(OpenSSL)
+  IF (${OpenSSL_FOUND})
+    MESSAGE(STATUS "Found OpenSSL on Darwin")
+  ELSE (${OpenSSL_FOUND})
+    MESSAGE(STATUS "Could not locate OpenSSL on Darwin. Setting OPENSSL_ROOT_DIR to /usr/local/opt/openssl")
+    set(OPENSSL_ROOT_DIR "/usr/local/opt/openssl")
+  ENDIF (${OpenSSL_FOUND})
+endif()
+
 find_package(OpenSSL 1.1 REQUIRED)
 
 ###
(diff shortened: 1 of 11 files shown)


CMakeLists.txt
```diff
@@ -1,8 +1,37 @@
 cmake_minimum_required(VERSION 3.12)
+set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake ${CMAKE_MODULE_PATH})
+
+
+###
+###  Semantic versioning
+###
+include ( GetGitRevisionDescription )
+#git_describe(VERSION)
```
bifurcation, line 9:
```suggestion
```

CMakeLists.txt
```diff
@@ -1,8 +1,37 @@
...
+git_get_exact_tag(VERSION)
+if(NOT VERSION)
+  message("No Git tag version found setting v0.0.0")
+  set(VERSION "v0.0.0")
+endif(NOT VERSION)
+
```
bifurcation, line 15:
```suggestion
```

CMakeLists.txt
```diff
@@ -1,8 +1,37 @@
...
+
+message( ${VERSION})
```
bifurcation, line 17:
```suggestion
message("Buiding version ${VERSION}")
```

CMakeLists.txt
```diff
@@ -1,8 +1,37 @@
...
+
```
bifurcation, line 18:
```suggestion
```

CMakeLists.txt
```diff
@@ -1,8 +1,37 @@
...
+
+get_git_head_revision(REFSPEC HASHVAR)
+
+string(REGEX REPLACE "^v([0-9]+)\\..*" "\\1" MLSPP_VERSION_MAJOR "${VERSION}")
+string(REGEX REPLACE "^v[0-9]+\\.([0-9]+).*" "\\1" MLSPP_VERSION_MINOR "${VERSION}")
+string(REGEX REPLACE "^v[0-9]+\\.[0-9]+\\.([0-9]+).*" "\\1" MLSPP_VERSION_PATCH "${VERSION}")
+string(REGEX REPLACE "^v[0-9]+\\.[0-9]+\\.[0-9]+(.*)" "\\1" MLSPP_VERSION_SHA1 "${VERSION}")
+mark_as_advanced(MLSPP_VERSION_MAJOR MLSPP_VERSION_MINOR MLSPP_VERSION_PATCH)
```
bifurcation, line 26: In the interest of making these CMakeLists reusable, I would avoid using `MLSPP` in variable names. Suggest `LIB_VERSION` etc.

CMakeLists.txt
```diff
@@ -52,6 +81,16 @@ endif()
 add_subdirectory(lib)
 
 # External libraries
+if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
```
bifurcation, line 84: Why is this branch necessary?  It seems like the below `find_package` line works fine with the macOS CI runners and `brew`.  Are you trying to avoid the need for the `brew reinstall` step in the CI run?

CMakeLists.txt
```diff
@@ -74,6 +113,28 @@ target_include_directories(${LIB_NAME}
...
 )
 
+###
+###  Library packaging and install
+###
+#include ( CPack )
+set ( dist_dir    ${CMAKE_BINARY_DIR}/dist )
+set ( prefix      ${CMAKE_INSTALL_PREFIX} )
+set ( exec_prefix ${CMAKE_INSTALL_PREFIX}/bin )
+set ( libdir      ${CMAKE_INSTALL_PREFIX}/lib )
```
bifurcation, line 123: I would prefer `lib_dir`, as with `dist_dir` above.  In any case, be consistent about whether there's an underscore.

CMakeLists.txt
```diff
@@ -74,6 +113,28 @@ target_include_directories(${LIB_NAME}
...
+set ( includedir  ${CMAKE_INSTALL_PREFIX}/include )
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/libmlspp.pc.in
+        ${CMAKE_CURRENT_BINARY_DIR}/libmlspp.pc @ONLY)
```
bifurcation, line 126:
```suggestion
               ${CMAKE_CURRENT_BINARY_DIR}/libmlspp.pc @ONLY)
```

README.md
```diff
@@ -31,3 +31,35 @@ Conventions
   * `X_key` is the public key of an asymmetric key pair
   * `X_priv` is the private key of an asymmetric key pair
   * `X_secret` is a symmetric secret
+
+Building, installing, linking etc
+----------------------------
```
bifurcation, line 36:
```suggestion
Installing and Uninstalling
---------------------------
```

README.md
```diff
@@ -31,3 +31,35 @@ Conventions
...
+> make install
+```
+
+There is no uninstall target, so use:
+```
+> xargs rm < install_manifest.txt
+```
+To get rid of most of the files (Some manual directory cleanup might be needed)
+
+Once installed there a pkgconfig file is installed so using this library from another cmake project could be done by:
```
bifurcation, line 57:
```suggestion
Once the library is installed, a pkgconfig file is also installed, so using this library from another cmake project could be done by:
```

bifurcation: Also, please wrap the line to 80 characters.

cmd/api_example/CMakeLists.txt
```diff
@@ -2,6 +2,6 @@ set(APP_NAME "api_example")
 
 file(GLOB APP_SOURCES CONFIGURE_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/*.cpp)
 
-add_executable(${APP_NAME} ${APP_SOURCES})
+add_executable(${APP_NAME} ${APP_SOURCES} ${version_file})
```
bifurcation, line 5: Just for consistency, could this be `VERSION_FILE`?

cmd/api_example/main.cpp
```diff
@@ -47,6 +49,9 @@ verify(const std::string& label, Session& alice, Session& bob)
 int
 main() // NOLINT(bugprone-exception-escape)
 {
+
+  std::cout << "Running library version:" << std::string(VERSION) << std::endl;
+  std::cout << "       git sha hash:" << std::string(HASHVAR) << std::endl;
```
bifurcation, line 54:
```suggestion
  std::cout << "Running library version: " << std::string(VERSION) << std::endl;
  std::cout << "           git sha hash: " << std::string(HASHVAR) << std::endl;
```

include/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+install ( DIRECTORY ../include/mls DESTINATION include
```
bifurcation, line 1: Can we move this to the main `CMakeLists.txt`?  

I would prefer not to have stuff in `include` other than `include/mls`.  Headers that are only needed internally should go in `src`.

lib/hpke/CMakeLists.txt
```diff
@@ -3,6 +3,15 @@ set(CURRENT_LIB_NAME hpke)
 ###
 ### Dependencies
 ###
+if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
```
bifurcation, line 6: Same comment here as above.

PR #139, opened 2020-11-04
Compiler errors

suhasHere:
Fixes few compiler errors that showed up on ``` make all ```

### DIFF
--- a/include/mls/key_schedule.h
+++ b/include/mls/key_schedule.h
@@ -44,7 +44,7 @@ struct SecretTree
   SecretTree() = default;
   SecretTree(CipherSuite suite_in,
              LeafCount group_size,
-             bytes application_secret_in);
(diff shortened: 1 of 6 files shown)


include/mls/key_schedule.h
```diff
@@ -44,7 +44,7 @@ struct SecretTree
   SecretTree() = default;
   SecretTree(CipherSuite suite_in,
              LeafCount group_size,
-             bytes encryption_secret);
+             bytes application_secret_in);
```
bifurcation, line 47: No, this should be `encryption_secret`.  Change it on the other side.

bifurcation (approved):

bifurcation:
Thanks for these fixes, but they should have gotten caught in CI.  Please file a follow-up to get clang-tidy re-enabled in CI.

PR #150, opened 2020-11-21
Make variant mapping more streamlined and compatible

bifurcation:
The variant selection introduced in #117 was nice in that it detached variant selection from the selected types, but it still required the awkward `::selector` construction.  And it turns out that some older compilers are unhappy with the template variable `type`, leading to undefined symbol errors if `type` is specialized in a .cpp file and duplicate symbol errors if specialized in .h.

This PR introduces an alternative approach based on an explicit `variant_map<Ts, Tv>` function, where `Ts` is the selector enum type and `Tv` is the mapped type.  This means that the mapping is detached from *both* the selector enum and the mapped type.

The main drawback is that the required boilerplate now needs to be defined in `namespace tls`.  So we end up with blocks of the following form at the bottom of some headers:

```C++
namespace tls {

using namespace mls;

TLS_VARIANT_MAP(CredentialType, BasicCredential, basic)
TLS_VARIANT_MAP(CredentialType, X509Credential, x509)

} // namespace tls
```

### DIFF
--- a/cmd/test_gen/main.cpp
+++ b/cmd/test_gen/main.cpp
@@ -371,7 +371,7 @@ generate_messages()
 
     // Construct an MLSCiphertext
     auto ciphertext = MLSCiphertext{
-      tv.group_id, tv.epoch,  ContentType::selector::application,
+      tv.group_id, tv.epoch,  ContentType::application,
       tv.random,   tv.random, tv.random,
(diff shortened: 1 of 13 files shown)


suhasHere (approved):
looks good to me ..

PR #151, opened 2020-11-25
X509 Credential API update

suhasHere:
X509Credential() takes vector<bytes>

### DIFF
--- a/src/credential.cpp
+++ b/src/credential.cpp
@@ -29,9 +29,22 @@ find_signature(Signature::ID id)
   throw InvalidParameterError("Unsupported algorithm");
 }
 
-X509Credential::X509Credential(
-  std::vector<X509Credential::CertData> der_chain_in)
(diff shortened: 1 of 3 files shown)


src/credential.cpp
```diff
@@ -29,9 +29,22 @@ find_signature(Signature::ID id)
   throw InvalidParameterError("Unsupported algorithm");
 }
 
-X509Credential::X509Credential(
-  std::vector<X509Credential::CertData> der_chain_in)
-  : der_chain(std::move(der_chain_in))
+static std::vector<mls::X509Credential::CertData>
+to_x509_credential_data(const std::vector<bytes>& data_in)
```
bifurcation, line 33:
```suggestion
bytes_to_x509_credential_data(const std::vector<bytes>& data_in)
```

bifurcation (approved):

PR #152, opened 2020-11-26
Expose Certificate attributes

suhasHere:


lib/hpke/src/certificate.cpp
```diff
@@ -25,14 +33,120 @@ struct Certificate::ParsedCertificate
...
+    // Clean up
+    sk_GENERAL_NAME_pop_free(san_names, GENERAL_NAME_free);
+    return san_info;
+  }
+
   explicit ParsedCertificate(X509* x509_in)
     : x509(x509_in, typed_delete<X509>)
     , sig_id(signature_algorithm(x509.get()))
+    , issuer(X509_NAME_oneline(X509_get_issuer_name(x509_in), nullptr, 0))
+    , subject(X509_NAME_oneline(X509_get_subject_name(x509_in), nullptr, 0))
```
bifurcation, line 136: This is broken.  It makes your issuer/subject matching depend on OpenSSL's stringification of distinguished names.  Instead, you want to use [the "canonical encoding" that is already provided](https://github.com/openssl/openssl/blob/a829b735b645516041b55746e013692babd8cd31/include/crypto/x509.h#L38).

lib/hpke/src/certificate.cpp
```diff
@@ -59,8 +180,23 @@ struct Certificate::ParsedCertificate
...
+      throw openssl_error();
+    }
+    return (0 != bc->ca);
+  }
+
   typed_unique_ptr<X509> x509;
   const Signature::ID sig_id;
+  const std::string issuer;
+  const std::string subject;
+  const std::string skID;
```
bifurcation, line 197: Let's spell these out, `subject_key_id` and `authority_key_id`.

lib/hpke/src/certificate.cpp
```diff
@@ -59,8 +180,23 @@ struct Certificate::ParsedCertificate
     return make_typed_unique<EVP_PKEY>(X509_get_pubkey(x509.get()));
   }
 
+  bool is_ca() const
```
bifurcation, line 183: If you're going to parse and store everything else at instantiation time, you might as well do this as well.

lib/hpke/src/openssl_common.cpp
```diff
@@ -63,6 +64,13 @@ typed_delete(X509* ptr)
   X509_free(ptr);
 }
 
+template<>
+void
+typed_delete(AUTHORITY_KEYID* ptr)
```
bifurcation, line 69: If you use the pre-parsed version on `X509`, you can delete this.

bifurcation (changes requested):

lib/hpke/src/certificate.cpp
```diff
@@ -25,14 +33,120 @@ struct Certificate::ParsedCertificate
...
+    }
+    san_names_nb = sk_GENERAL_NAME_num(san_names);
+
+    // Check each name within the extension
+    for (int i = 0; i < san_names_nb; i++) {
+      const GENERAL_NAME* current_name = sk_GENERAL_NAME_value(san_names, i);
+
+      if (current_name->type == GEN_DNS) {
+        // NOLINTNEXTLINE (cppcoreguidelines-pro-type-cstyle-cast)
+        const char* dns_name = (const char*)(ASN1_STRING_get0_data(
```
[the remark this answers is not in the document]
suhasHere: i did that , it ended up with a diferent clang error and hence decided to go this route

lib/hpke/src/certificate.cpp
suhasHere: might need some help on the right accessor on this

lib/hpke/src/certificate.cpp
```diff
@@ -25,14 +39,104 @@ struct Certificate::ParsedCertificate
...
+    }
+
+    return san_info;
+  }
+
   explicit ParsedCertificate(X509* x509_in)
     : x509(x509_in, typed_delete<X509>)
     , sig_id(signature_algorithm(x509.get()))
+    , issuer(X509_NAME_oneline(X509_get_issuer_name(x509.get()), nullptr, 0))
+    , subject(X509_NAME_oneline(X509_get_subject_name(x509.get()), nullptr, 0))
```
bifurcation, line 124: These are still wrong.  My guess is that you want one of two patterns here:

1. `X509_NAME_get0_der(X509_get_subject_name(x509))`
2. `X509_subject_name_hash(x509)`

I think you want the latter, since you're only using the subject and issuer for comparison.  That means that `issue` and `subject` can just be long integers (let's use uint64_t to be explicit).

lib/hpke/src/certificate.cpp
```diff
@@ -25,14 +39,104 @@ struct Certificate::ParsedCertificate
...
+    }
+    const auto* octet_akid = ASN1_STRING_get0_data(asn_akid);
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+    std::string akid_str(reinterpret_cast<const char*>(octet_akid));
+    return bytes(akid_str.begin(), akid_str.end());
+  }
+
+  static std::vector<GeneralName> parse_san(X509* cert)
+  {
+    std::vector<GeneralName> san_info;
```
bifurcation, line 70: Nit: Something like `names` would be clearer.

lib/hpke/src/certificate.cpp
```diff
@@ -25,14 +39,104 @@ struct Certificate::ParsedCertificate
...
+  // Parse Subject Key Identifier Extension
+  static bytes parse_skid(X509* cert)
+  {
+    const auto* asn_skid = X509_get0_subject_key_id(cert);
+    if (asn_skid == nullptr) {
+      return bytes{};
+    }
+    const auto* octet_skid = ASN1_STRING_get0_data(asn_skid);
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+    std::string skid_str(reinterpret_cast<const char*>(octet_skid));
```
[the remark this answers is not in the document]
suhasHere: good point... will update

lib/hpke/src/certificate.cpp
```diff
@@ -25,14 +39,104 @@ struct Certificate::ParsedCertificate
...
+    return bytes(skid_str.begin(), skid_str.end());
+  }
+
+  // Parse Authority Key Identifier
+  static bytes parse_akid(X509* cert)
```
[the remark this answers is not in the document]
suhasHere: +1

lib/hpke/src/certificate.cpp
```diff
@@ -1,16 +1,59 @@
+#include "group.h"
+#include "openssl_common.h"
+#include <cstring>
 #include <hpke/certificate.h>
 #include <hpke/signature.h>
+#include <iostream>
```
bifurcation, line 6:
```suggestion
```

lib/hpke/src/certificate.cpp
```diff
@@ -1,16 +1,59 @@
...
 #include <openssl/x509.h>
+#include <openssl/x509v3.h>
+#include <tls/compat.h>
+namespace hpke {
```
bifurcation, line 10:
```suggestion

namespace hpke {
```

lib/hpke/src/certificate.cpp
```diff
@@ -1,16 +1,59 @@
...
 
-namespace hpke {
+static std::string
+asn1_string_to_std_string(const asn1_string_st* asn1_string)
+{
+  const auto* data = ASN1_STRING_get0_data(asn1_string);
+  const auto data_size = static_cast<size_t>(ASN1_STRING_length(asn1_string));
+  auto str =
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+    std::string(reinterpret_cast<const char*>(data));
```
bifurcation, line 34:
```suggestion
  // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
  auto str = std::string(reinterpret_cast<const char*>(data));
```

lib/hpke/src/certificate.cpp
```diff
@@ -1,16 +1,59 @@
...
+  if (str.size() != data_size) {
+    throw std::runtime_error("Malformed certificate");
```
bifurcation, line 36:
```suggestion
    throw std::runtime_error("Malformed ASN.1 string");
```

lib/hpke/src/certificate.cpp
```diff
@@ -1,16 +1,59 @@
...
+  }
+  const auto* ptr = ASN1_STRING_get0_data(octets);
+  const auto len = ASN1_STRING_length(octets);
+  // NOLINTNEXTLINE (cppcoreguidelines-pro-bounds-pointer-arithmetic)
+  return bytes(ptr, ptr + len);
+}
 
-namespace hpke {
+static std::string
+asn1_string_to_std_string(const asn1_string_st* asn1_string)
```
bifurcation, line 28:
```suggestion
asn1_string_to_std_string(const ASN1_STRING* asn1_string)
```

bifurcation (approved):
Just a few nits.

PR #159, opened 2021-01-04
use vcpkg instead of brew to handle dependencies

dimarusyy:
- added vcpkg as submodule
- added vcpkg.json
- updated CI files


dimarusyy:
@bifurcation , please let me know if you're ok with these changes and keep `vcpkg` as submodule.
Also, I'm not sure I understand correctly why pkgconfig is needed.

bifurcation:
Thanks for taking a look at this, @dimarusyy.  Some comments:

1. It seems like the main benefit to vcpkg here is to make sure it works on Windows, right?  Assuming that's the case, please add a Windows build to `build.yml`.  If there are other issues (e.g., I suspect clang-tidy will be an issue), we can put that off.

2. vcpkg looks to be noticeably slower than `brew` on the platforms we have (26min for macOS!).  Would it be possible to use vcpkg only where necessary? (which probably means only on Windows)  In code, this would probably manifest as a CMake option to enable vcpkg, as for clang-tidy.

3. I have generally preferred fetching things at configure time (`cmake` time) as opposed to git submodules.  That would still be my preference here if it were feasible.

With regard to (2) and (3), [this StackOverflow thread](https://stackoverflow.com/questions/62476438/configure-and-install-required-packages-for-a-project) and [this template repo](https://github.com/codusnocturnus/cmake-vcpkg-template) might be helpful.

dimarusyy:
@bifurcation , please see my comments:
1. I was planning to add windows builds in a different PR. But I'm to add it in this PR.
2. Yes, it is slower. The solution is to cache prebuild binaries with `vcpkg` but I don't know if this is possible with github CI as it requires some storage to keep prebuild packages. I'll update if I find a good solution for that.
3. Using `vcpkg` as submodule, it's native cmake integration and specifying list of dependencies in `vcpkg.json` file is the way recommended by authors of this package manager. Besides, invoking external commands to install packages with vcpkg doesn't make sense for me. That's why I would not move on to this direction.

Taking into account your comments I don't think `vcpkg` could be used now in github CI.

dimarusyy:
@bifurcation , github images have vcpkg already installed. I updated CI configs to use pre-installed vcpkg and cache prebuild directory with dependencies. Also, I have updated cmakefile's to allow invoking ctest. 
Please, review. 
PS : there are quite many commits and maybe I should create the new PR.

.github/workflows/build.yml
```diff
@@ -10,40 +10,54 @@ on:
...
 
     - name: build
-      run: make everything
-
-    - name: run tests
-      run: make test-all
+      run: |
+        cmake --build "${{ env.CMAKE_BUILD_DIR }}"
 
+    - name: Unit tests
```
bifurcation, line 61: This is not running the tests on the libraries right now, but we can handle that in a follow-up.

.github/workflows/compat.yml
```diff
@@ -10,22 +10,32 @@ on:
...
+      uses: actions/cache@v2
+      with:
+        path: |
+            ${{ env.CMAKE_BUILD_DIR }}/vcpkg_installed
+        key: ${{ runner.os }}-${{ hashFiles( '**/vcpkg.json' ) }}
+
     - name: build the library
-      run: make
+      run: |
+        cmake -B "${{ env.CMAKE_BUILD_DIR }}" -DCMAKE_BUILD_TYPE=Debug -DCMAKE_TOOLCHAIN_FILE="$VCPKG_INSTALLATION_ROOT/scripts/buildsystems/vcpkg.cmake" .
```
bifurcation, line 40: Nit: Just to shorten this line a bit, it would be nice to push the `CMAKE_TOOLCHAIN_FILE` value to an environment variable and use the variable here.

CMakeLists.txt
```diff
@@ -97,7 +97,7 @@ target_include_directories(${LIB_NAME}
 ###
 ### Tests
 ###
-
+enable_testing()
```
bifurcation, line 100: Why is this change needed?  It seems problematic, since with this change, `enable_testing()` is not called before `lib/hpke/test/CMakeLists.txt` is processed (and the other library test files).  I would prefer to keep the `enable_testing()` directives in the test directories, where they are now.

lib/hpke/test/CMakeLists.txt
```diff
@@ -12,5 +12,5 @@ target_link_libraries(${TEST_APP_NAME} ${CURRENT_LIB_NAME} bytes doctest::doctes
 
 # Enable CTest
 include(doctest)
-enable_testing()
+add_test(${TEST_APP_NAME} ${TEST_APP_NAME})
```
bifurcation, line 15: This change seems duplicative with the `doctest_discover_tests()` call below.  Why is it needed?

bifurcation (commented):
This is looking pretty good.  Just a few quick questions, then I think it's ready to merge.

bifurcation:
Also, don't worry about the number of commits, we can squash and merge anyway.

CMakeLists.txt
dimarusyy: Maybe `enable_testing()` should be moved before first `add_diretcory()` call. My idea was to have a separate target that will invoke running unit-tests instead of moving to a separate forlder and running `ctest` there.

dimarusyy (commented):
@bifurcation , I have moved `enable_testing()` up in the main CMakeLists.txt and removed `add_test(...)` as you suggested as that was really useless.

bifurcation:
We might want to tweak the build config going forward, but this seems close enough to merge.  Thanks for putting this together @dimarusyy!  I think you actually managed to speed up the CI builds!

PR #160, opened 2021-01-04
Tie in to the interop harness

bifurcation:
This PR adds an executable that ties into the gRPC-based interop test framework.  Right now, the focus is on test vectors.  As the RPC interface expands to include MLS behaviors, further work will be needed, but should be of the same general flavor -- mapping gRPC calls to mlspp library calls.


dimarusyy (commented):
Richard, did you consider building some abstraction for `interop` instead of focusing only on grpc ? boost.beast build on top of boost.asio is a good example.

bifurcation:
@dimarusyy Better to ask that question over in the [MLS implementations](https://github.com/mlswg/mls-implementations) repo and associated Wire room.  The consensus so far has been that gRPC is ok since it avoids the need to bother with a bunch of HTTP API details.

dimarusyy:
> @dimarusyy Better to ask that question over in the [MLS implementations](https://github.com/mlswg/mls-implementations) repo and associated Wire room. The consensus so far has been that gRPC is ok since it avoids the need to bother with a bunch of HTTP API details.

I'm not sure how can I join with non-guest account to wire and ask...and whether it makes sense as consensus reached.
IMHO,  having some kind of network abstraction without network protocol details is a good option. Please, consider an example when signaling related to MLS protocol is done via UDP (RTCP protocol).
Besides, not all users would be happy to compile protobuf/grpc code if not needed.

bifurcation:
Oh, to be clear, the interop code here is only for interop testing!  Users of the library don't have to care about all of this gRPC stuff.  They're free to send the messages however they like.

suhasHere (approved):

dimarusyy (approved):

PR #161, opened 2021-01-06
Add RSA signatures

bifurcation:
Doesn't pass tests or do serialization yet.


lib/hpke/include/hpke/signature.h
```diff
@@ -16,6 +16,7 @@ struct Signature
     P521_SHA512,
     Ed25519,
     Ed448,
+    RSA_SHA256,
```
pabuhler, line 19: The  WCA used sha512WithRSAEncryption, do you need to add explicit support for 512?

lib/hpke/src/signature.cpp
```diff
@@ -95,69 +101,246 @@ struct ConcreteSignature : public Signature
...
+        throw openssl_error();
+      }
+      return std::make_unique<PublicKey>(pkey.get());
+    }
+
+    typed_unique_ptr<EVP_PKEY> pkey;
+  };
+
+  explicit RSASignature(Digest::ID digest)
+    : Signature(Signature::ID::RSA_SHA256)
```
bifurcation, line 144: Let's be hash-agnostic here.
```suggestion
    : Signature(digest_to_sig(digest))
```

... where `digest_to_sig()` should just be a `switch` statement like `digest_to_md` -- but with only one arm, for SHA256.

lib/hpke/src/signature.cpp
```diff
@@ -95,69 +101,246 @@ struct ConcreteSignature : public Signature
...
+
+  static const EVP_MD* digest_to_md(Digest::ID digest)
+  {
+    switch (digest) {
+      case Digest::ID::SHA256:
+        return EVP_sha256();
+      case Digest::ID::SHA384:
+        return EVP_sha384();
+      case Digest::ID::SHA512:
+        return EVP_sha512();
```
bifurcation, line 291:
```suggestion
```

lib/hpke/src/signature.cpp
```diff
@@ -95,69 +101,246 @@ struct ConcreteSignature : public Signature
...
+      throw openssl_error();
+    }
+
+    if (1 !=
+        EVP_DigestSignInit(ctx.get(), nullptr, md, nullptr, rsk.pkey.get())) {
+      throw openssl_error();
+    }
+
+    static const size_t max_sig_size = 256;
+    auto siglen = max_sig_size;
```
bifurcation, line 247:
```suggestion
    auto siglen = EVP_PKEY_size(rsk.pkey.get());
```

Also, please make the corresponding change for the group-based signatures.

bifurcation (commented):
Please also add support to `Certificate`.  You'll need to change the [public key ctors](https://github.com/cisco/mlspp/blob/master/lib/hpke/src/certificate.cpp#L196) to be smarter, and switch between `GroupSignature::PublicKey` and `RSASignature::PublicKey` as appropriate.

bifurcation (commented):
Thanks for finishing this one, @suhasHere.  GitHub won't let me approve the PR, but it has my approval.

PR #162, opened 2021-01-07
Add Logging

bifurcation:
This PR adds a basic logging facility and installs a few basic crypto logs.  It does logging only at the MLS level.  The libraries (TLS syntax, HPKE) should have their own validation.  More work to do to add appropriate logging throughout.


dimarusyy (commented):
Richard, if you have plans to switch to `vcpkg` using spdlog could be an option to what you've implemented as it supports both runtime and compile time log level verification.

Also, I think having a list of sinks in `Log` struct would be a good idea.
Say, adding sink for specific subsystem (maybe with some specific tag) and then invoking it via `Log::error<key_schedule_>("some message")` would be much better for readability.

bifurcation:
I had considered bringing in a dependency like spdlog, but for something like this that's going to get integrated into a bigger application, I think it's better to have a thin layer that the **application** could plug in to something like spdlog.  In other words, the library here doesn't need logging, it needs an collector that will pass log lines to a logging system.

In a similar vein, I don't think we should dive into the complexities of multiple collectors.  The application can do that as part of its `log::Sink` implementation.

dimarusyy (approved):

src/crypto.cpp
```diff
@@ -10,6 +11,9 @@ using hpke::KDF;       // NOLINT(misc-unused-using-decls)
 using hpke::KEM;       // NOLINT(misc-unused-using-decls)
 using hpke::Signature; // NOLINT(misc-unused-using-decls)
 
+using mls::log::Log;
+static const auto log_mod = "crypto"s;
+
```
suhasHere, line 16: why do we need log_mod to be a string_literal instead of normal string

suhasHere (approved):

src/crypto.cpp
bifurcation: I think it could also be a std::string, this is just syntactic sugar.

PR #166, opened 2021-01-15
fix signature scheme parse

suhasHere:
Current cert sets cert's signature algo instead of public key algo. This pr fixes that issue. Thanks @pabuhler  for the unit test

### DIFF
--- a/lib/hpke/src/certificate.cpp
+++ b/lib/hpke/src/certificate.cpp
@@ -92,6 +92,7 @@ struct Certificate::ParsedCertificate
     return asn1_octet_string_to_bytes(X509_get0_authority_key_id(cert));
   }
 
+  // Parse Subject Alternate Names
   static std::vector<GeneralName> parse_san(X509* cert)
   {
     std::vector<GeneralName> names;
@@ -125,7 +126,6 @@ struct Certificate::ParsedCertificate
 
   explicit ParsedCertificate(X509* x509_in)
     : x509(x509_in, typed_delete<X509>)
-    , sig_id(signature_algorithm(x509.get()))
     , issuer(X509_issuer_name_hash(x509.get()))
     , subject(X509_subject_name_hash(x509.get()))
     , subject_key_id(parse_skid(x509.get()))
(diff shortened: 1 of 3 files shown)


lib/hpke/src/certificate.cpp
```diff
@@ -193,28 +170,62 @@ struct Certificate::ParsedCertificate
...
+    case EVP_PKEY_ED448:
+      return Signature::ID::Ed448;
+    case EVP_PKEY_EC: {
+      auto key_size = EVP_PKEY_bits(pkey);
+      return (key_size == 256) ? Signature::ID::P256_SHA256
+                               : (key_size == 384) ? Signature::ID::P384_SHA384
+                                                   : Signature::ID::P521_SHA512;
+    }
+    case EVP_PKEY_RSA:
+      return Signature::ID::RSA_SHA256;
```
bifurcation, line 191:
```suggestion
      // It is not necessarily the case that RSA keys are used with SHA256, 
      // but in practically all cases, it is.  Since the SubjectPublicKeyInfo
      // key doesn't tell us any more information, we assume this is the case.
      return Signature::ID::RSA_SHA256;
```

test/session.cpp
```diff
@@ -225,3 +225,87 @@ TEST_CASE_FIXTURE(RunningSessionTest, "Full Session Life-Cycle")
     check(initial_epoch);
   }
 }
+
+TEST_CASE("Session with X509 Credential")
```
bifurcation, line 229: It seems like it would be good to test with an RSA public key in the leaf.  Would it be straightforward to generate that test data?

suhasHere: hpke/test/certificate.cpp has a test with rsa all through..

bifurcation: That just tests *parsing*.  The bug being fixed here is that the public key in the certificate couldn't be used to verify anything.  So that's what you need to test.  You're probably correct that it should be in `lib/hpke/test/certificate.cpp`.  If you don't want to mess with importing private keys, you could just have your test generation script produce an example (message, signature) under the certified public key, and verify that the parsed public key can be used to verify the signature.

bifurcation:
Seems like this was fixed with the revised approach #169

PR #169, opened 2021-01-26
support RSA sha1 cert as sha256

suhasHere:


suhasHere:
this looks good . can't approve my self, but it is good to merge. thanks

PR #171, opened 2021-01-30
JSON-based interop testing

bifurcation:
Corresponds to https://github.com/mlswg/mls-implementations/pull/33

### DIFF
--- a/cmd/interop/src/json_details.h
+++ b/cmd/interop/src/json_details.h
@@ -96,6 +96,10 @@ struct tls_serializer
   {};
 
 TLS_SERIALIZER(mls::HPKEPublicKey)
+TLS_SERIALIZER(mls::TreeKEMPublicKey)
+TLS_SERIALIZER(mls::MLSPlaintext)
+TLS_SERIALIZER(mls::UpdatePath)
+TLS_SERIALIZER(mls::KeyPackage)
 
 } // namespace nlohmann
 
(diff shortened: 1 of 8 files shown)


suhasHere (approved):

cmd/interop/src/json_details.h
```diff
@@ -118,4 +126,79 @@ NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE(EncryptionTestVector,
                                    sender_data_info,
                                    leaves);
 
+NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE(KeyScheduleTestVector::Epoch,
```
dimarusyy, line 129: I would seriously consider switching to boost::json

bifurcation: Any particular reason?

dimarusyy: Mainly performance.

bifurcation: Since this is just for interop testing, I'm not really worried about performance.  It also looks like `boost::json` lacks any sort of easy type conversion of the kind we have here.  So I'm going to pass for now.

dimarusyy: `boost;:json` uses a different approach : http://vinniefalco.github.io/doc/json/json/usage/conversion.html.

PR #174, opened 2021-02-02
Test vector bugs

bifurcation:
A couple of bugs that I noticed while looking at test vectors:

* TreeKEM was not properly setting `update_secret` on `TreeKEMPrivateKey`
* Nonces were being derived to `key_size` not `nonce_size`
* Key schedule test vectors were capturing behavior specific to mlspp (we initialize a whole key schedule epoch for the one-member group, even though this is not visible to other participants)


suhasHere (approved):
wow .. nice find .. LGTM

PR #175, opened 2021-02-05
SFrame parameter negotiation

bifurcation:
Implements SFrame parameter negotiation as specified in the [corresponding spec PR](https://github.com/bifurcation/sframe-mls/pull/3).  

Note that this is the first extension we have that is copied into the group state.  Currently application-defined extensions cannot be auto-copied over from the GroupInfo to the GroupContext.  The application would have to manually call `state.extensions().add(custom_extension)`, and even that is currently impossible because the `extensions()` method is `const`.  We might want to consider having a better story for application-defined extensions.

### DIFF
--- a/test/state.cpp
+++ b/test/state.cpp
@@ -97,7 +97,7 @@ TEST_CASE_FIXTURE(StateTest, "Two Person")
 TEST_CASE_FIXTURE(StateTest, "SFrame Parameter Negotiation")
 {
   // Set the SFrame parameters for the group
-  auto specified_params = SFrameParameters{ 1, 4 };
+  auto specified_params = SFrameParameters{ 1, 8 };
   auto group_extensions = ExtensionList{};
   group_extensions.add(specified_params);
 
@@ -119,7 +119,7 @@ TEST_CASE_FIXTURE(StateTest, "SFrame Parameter Negotiation")
 
   auto decoded_capabilities =
     key_packages[1].extensions.find<SFrameCapabilities>();
-  REQUIRE(!!decoded_capabilities);
+  REQUIRE(decoded_capabilities);
   REQUIRE(opt::get(decoded_capabilities) == compatible_capabilities);
   REQUIRE(opt::get(decoded_capabilities).compatible(specified_params));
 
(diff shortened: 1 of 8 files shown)


test/state.cpp
```diff
@@ -94,11 +94,65 @@ TEST_CASE_FIXTURE(StateTest, "Two Person")
...
+  // Create the initial state of the group
+  auto first0 = State{ group_id,          suite,           init_privs[0],
+                       identity_privs[0], key_packages[0], group_extensions };
+
+  // Get a KeyPackage from the second member, and verify compatiblity
+  auto compatible_capabilities = SFrameCapabilities{ { 1, 2, 3, 4 } };
+  REQUIRE(compatible_capabilities.compatible(specified_params));
+
+  auto incompatible_capabilities = SFrameCapabilities{ { 2, 3, 4 } };
+  REQUIRE_FALSE(incompatible_capabilities.compatible(specified_params));
```
suhasHere, line 113: isn't 4 common between the lists here..

test/state.cpp
```diff
@@ -94,11 +94,65 @@ TEST_CASE_FIXTURE(StateTest, "Two Person")
...
+
+  auto key_package_extensions = ExtensionList{};
+  key_package_extensions.add(compatible_capabilities);
+  key_packages[1].sign(identity_privs[1],
+                       KeyPackageOpts{ key_package_extensions });
+
+  auto decoded_capabilities =
+    key_packages[1].extensions.find<SFrameCapabilities>();
+  REQUIRE(!!decoded_capabilities);
```
suhasHere, line 122: REQUIRE_TRUE

test/state.cpp
```diff
@@ -94,11 +94,65 @@ TEST_CASE_FIXTURE(StateTest, "Two Person")
...
+
+  auto group = std::vector<State>{ first1, second0 };
+  verify_group_functionality(group);
+
+  // Check that both participants have the  correct SFrame parameters
+  auto first_params = first1.extensions().find<SFrameParameters>();
+  auto second_params = second0.extensions().find<SFrameParameters>();
+  REQUIRE(!!first_params);
+  REQUIRE(!!second_params);
+  REQUIRE(opt::get(first_params) == specified_params);
```
suhasHere, line 143: REQUIRE_TRUE

suhasHere (commented):
one clarification question in the test. Otherwise looks good to me

test/state.cpp
bifurcation: No, `4` is the value for `epoch_bits`.  I'll change it to be clearer, though.

suhasHere (approved):

PR #177, opened 2021-02-09
Add Subject Parsing

suhasHere:
Cert parser returns certificate subject as hash output. Some applications might need access to the actual entry for verification purposes. This PR parses subject CN into string. It also adds support for parsing Subject organization


lib/hpke/src/certificate.cpp
```diff
@@ -80,6 +80,26 @@ struct Certificate::ParsedCertificate
     return digest;
   }
 
+  // Parse Subject CN into textual representation
+  static std::string parse_subject_name(X509* cert)
+  {
+    auto x509_name = X509_get_subject_name(cert);
```
dimarusyy, line 86: auto *x509_name ?

dimarusyy (approved):

lib/hpke/src/certificate.cpp
```diff
@@ -54,6 +54,12 @@ struct DNSName
 
 using GeneralName = tls::var::variant<RFC822Name, DNSName>;
 
+struct ParsedX509Subject
```
bifurcation, line 57: It's pretty gross to just handle some specific fields out of the `Name` struct.  How about something like the following:

```C++
enum struct NameType : int {
  organization = NID_organizationName,
  organizational_unit = NID_organizationalUnitName,
  country = NID_countryName,
  state_or_province_name = NID_stateOrProvinceName,
  common_name = NID_commonName,
  serial_number = NID_serialNumber,
};

using ParsedName = std::map<NameType, std::string>;
```

That way your parsing code can just set `name[nid] = parsed_value`.

bifurcation: And for completeness, might as well do this for the issuer as well.

lib/hpke/src/certificate.cpp
```diff
@@ -80,6 +86,45 @@ struct Certificate::ParsedCertificate
     return digest;
   }
 
+  // Parse Subject fields (CN, O=)
+  static ParsedX509Subject parse_subject(X509* cert)
```
bifurcation, line 90: This should work on an `const X509_NAME*` so that you can use it with both subject and issuer.

bifurcation: Also, please add a comment here noting that this method does not implement fully general name parsing.  Duplicate entries are not supported; if they are present, the last one presented by OpenSSL is chosen.

lib/hpke/src/certificate.cpp
```diff
@@ -246,6 +294,18 @@ Certificate::subject() const
   return parsed_cert->subject;
 }
 
+std::string
+Certificate::subject_common_name() const
```
bifurcation, line 298: Let's not destructure like this; instead just return the `ParsedName` representing the subject / issuer.

bifurcation (changes requested):

bifurcation (approved):

PR #181, opened 2021-02-21
Remove ExternalProject dependency

bifurcation:
This speeds up builds and lets consumers build without needing to fetch the dependency.  The `mpark::variant` library hasn't been updated in 2 years anyway, so it seems unlikely we'll fall behind.


dimarusyy:
@bifurcation , does it make sense to avoid using mpark.variant when building for c++17 ? 
Btw, does something stops from supporting only C++17 and newer standard ?

bifurcation:
@dimarusyy - The issue is for consumers that want to be able to [deploy on macOS back to 10.11](https://github.com/cisco/mlspp/blob/main/.github/workflows/compat.yml).  The 10.11 target supports `std::variant` at a superficial level, but [not anything that would throw `std::bad_variant_access`](https://stackoverflow.com/questions/52521388/stdvariantget-does-not-compile-with-apple-llvm-10-0), such as `std::get<T>()`.

But it's a good point that we only need to use this library when that compatibility is required.  I'll think about how to add a switch.

dimarusyy (approved):

Issue #190, opened 2021-03-16
Randomly broken MLSCiphertext


franziskuskiefer:
Running the interop test vectors sometimes the MLSCiphertext looks like this

```
 [4, 0, 1, 2, 3, 0, 1, 2, 3, 4, 5, 6, 7, 1, 1c, ae, 5, dc, 8b, 5c, 4c, ...
```

which translates to
* group id: `[0, 1, 2, 3]`
* epoch: `0x0001020304050607`
* content_type: `Application`

The next element should be the authenticated data vector of length up to 2^32. But it's `1c, ae, 5, dc, 8b, 5c, 4c, ...`, which can't be correct. This looks like maybe the length header is missing.
Note that other times this looks correct, e.g. ` [0, 0, 0, 4, 1, 2, 3, 4, 1C, DE, 32, ...`.

bifurcation:
Looks like the `encrypted_sender_data` and `authenticated_data` fields are swapped in [the encode/decode specification](https://github.com/cisco/mlspp/blob/main/include/mls/messages.h#L545).  Should be a quick fix.

franziskuskiefer:
#191 has fixed this issue, thanks!

PR #191, opened 2021-03-16
Correct the order of fields in MLSCiphertext

bifurcation:
As @franziskuskiefer noted in #190, MLSCiphertext objects produced by mlspp do not comply with the spec w.r.t. field order.  In particular, `encrypted_sender_data` is serialized before `authenticated_data`.  This PR aligns MLSCiphersuite to the specification.

### DIFF
--- a/include/mls/messages.h
+++ b/include/mls/messages.h
@@ -528,8 +528,9 @@ private:
 
 // struct {
 //     opaque group_id<0..255>;
-//     uint32 epoch;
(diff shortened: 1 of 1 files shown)


franziskuskiefer (approved):
Thanks! This could be the issue. I can test tomorrow if it fixes the issues I see. But it's a good patch in any case.

suhasHere (approved):

PR #193, opened 2021-03-18
remove using declaratives

suhasHere:
Removes ``` using namespace mls``` directive from includes under src. The ones under cmd/ test/ andd mls_vectors/ are untouched.

### DIFF
--- a/include/mls/credential.h
+++ b/include/mls/credential.h
@@ -109,9 +109,7 @@ private:
 
 namespace tls {
 
-using namespace mls;
-
-TLS_VARIANT_MAP(CredentialType, BasicCredential, basic)
-TLS_VARIANT_MAP(CredentialType, X509Credential, x509)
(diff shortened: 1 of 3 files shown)


include/mls/credential.h
```diff
@@ -109,9 +109,7 @@ class Credential
 
 namespace tls {
 
-using namespace mls;
-
-TLS_VARIANT_MAP(CredentialType, BasicCredential, basic)
-TLS_VARIANT_MAP(CredentialType, X509Credential, x509)
+TLS_VARIANT_MAP(mls::CredentialType, mls::BasicCredential, basic)
+TLS_VARIANT_MAP(mls::CredentialType, mls::X509Credential, x509)
```
bifurcation, line 113: Why are the `mls::` prefixes required here, but not in `messages.h`?

suhasHere: compiler reports error on only this. other it was throwing error if i use mls::

bifurcation (approved):

PR #194, opened 2021-03-19
Update HPKE implementation to current spec

bifurcation:
This PR implements a few changes in recent versions of HPKE:

* `HPKE-05 -> HPKE-v1`
* Updates to the way `secret` is computed in the key schedule
* Implements the export-only pseudo-AEAD
* Fixes a bug where ECDH private keys are not fixed size

With these changes, our HPKE implementation passes the current test vectors (though we do not test all of the encryptions in order to keep the test vector build time down).


lib/hpke/include/hpke/hpke.h
```diff
@@ -114,6 +114,7 @@ struct AEAD
     AES_128_GCM = 0x0001,
     AES_256_GCM = 0x0002,
     CHACHA20_POLY1305 = 0x0003,
+    export_only = 0xffff,
   };
```
suhasHere, line 118: wondering one line comment might be handy on export_only ?

suhasHere (approved):

lib/hpke/src/hpke.cpp
```diff
@@ -235,6 +228,14 @@ AEAD::get<AEAD::ID::CHACHA20_POLY1305>()
   return AEADCipher::get<AEAD::ID::CHACHA20_POLY1305>();
 }
 
+template<>
+const AEAD&
+AEAD::get<AEAD::ID::export_only>()
+{
+  return ExportOnlyCipher::get();
```
dimarusyy, line 235: What about creating static object here instead of making ExportOnlyCipher as singleton ?

bifurcation: Do you mean you would prefer if the singleton were a static member of `ExportOnlyCipher`, like with `::instance` on other types here?

dimarusyy: I mean something like this :

> template<>
> const AEAD&
> AEAD::get<AEAD::ID::export_only>()
> {
>     static ExportOnlyCipher eoc;
>     return eoc;
> }

and make ExportOnlyCipher not a singleton class.

dimarusyy (approved):

PR #197, opened 2021-03-25
External tree

bifurcation:
Right now, we require that the Welcome/GroupInfo include a ratchet_tree extension containing the group's ratchet tree.  This PR makes that extension optional, on both the committer and joiner sides.

### DIFF
--- a/include/mls/state.h
+++ b/include/mls/state.h
@@ -89,7 +89,8 @@ public:
   MLSPlaintext remove(RosterIndex index) const;
   MLSPlaintext remove(LeafIndex removed) const;
 
-  struct CommitOpts {
+  struct CommitOpts
+  {
     std::vector<Proposal> extra_proposals;
     bool inline_tree;
   };
--- a/include/mls/state.h
+++ b/include/mls/state.h
@@ -89,9 +89,13 @@ public:
   MLSPlaintext remove(RosterIndex index) const;
   MLSPlaintext remove(LeafIndex removed) const;
 
+  struct CommitOpts {
+    std::vector<Proposal> extra_proposals;
+    bool inline_tree;
+  };
   std::tuple<MLSPlaintext, Welcome, State> commit(
     const bytes& leaf_secret,
-    const std::vector<Proposal>& extra_proposals) const;
+    const std::optional<CommitOpts>& opts) const;
 
   ///
   /// Generic handshake message handler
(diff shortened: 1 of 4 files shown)


include/mls/state.h
```diff
@@ -88,9 +89,14 @@ class State
   MLSPlaintext remove(RosterIndex index) const;
   MLSPlaintext remove(LeafIndex removed) const;
 
+  struct CommitOpts
```
bifurcation, line 92: This should probably move to the top level, not scoped within `State`.

src/state.cpp
```diff
@@ -161,8 +165,9 @@ State::external_join(const bytes& leaf_secret,
 {
   auto initial_state = State(std::move(sig_priv), pgs);
   auto add = initial_state.add_proposal(kp);
+  auto opts = CommitOpts{ { add }, true };
```
bifurcation, line 168:
```suggestion
  auto opts = CommitOpts{ { add }, false };
```

Nobody is going to use the Welcome here, so we shouldn't incur the expense of serializing out the tree.

test/state.cpp
```diff
@@ -82,12 +82,38 @@ TEST_CASE_FIXTURE(StateTest, "Two Person")
...
+  // Handle the Add proposal and create a Commit
+  auto add = first0.add_proposal(key_packages[1]);
+  // Don't generate RatchetTree extension
+  auto opts = State::CommitOpts{ { add }, false };
+  auto [commit, welcome, first1] = first0.commit(fresh_secret(), opts);
+  silence_unused(commit);
+
+  // Initialize the second participant from the Welcome, pass in the
+  // tree externally
+  auto second0 = State{
```
bifurcation, line 114: Suggest adding another check before this of the form to verify that the State ctor throws when there's no tree:
```
CHECK_THROWS(State{..., std::nullopt});
```

test/state.cpp
```diff
@@ -82,12 +82,38 @@ TEST_CASE_FIXTURE(StateTest, "Two Person")
 
   // Handle the Add proposal and create a Commit
   auto add = first0.add_proposal(key_packages[1]);
-  auto [commit, welcome, first1] = first0.commit(fresh_secret(), { add });
+  auto opts = State::CommitOpts{ { add }, true };
```
bifurcation, line 85: It would be nicer if we could put these back inline.  Maybe moving `CommitOpts` to the top level will make it compact enough?

bifurcation (commented):
I can't approve this because I started it, but this LGTM with a couple of minor things.

PR #202, opened 2021-04-21
Enable FIPS compatibility

bifurcation:
When using FIPS-approved algorithms, MLSpp should be compatible with FIPS-enabled crypto libraries.  This PR enables FIPS testing and works around one minor incompatibility.

### DIFF
--- a/lib/hpke/src/digest.cpp
+++ b/lib/hpke/src/digest.cpp
@@ -87,4 +87,46 @@ Digest::hmac(const bytes& key, const bytes& data) const
   return md;
 }
 
+bytes
+Digest::hmac_for_hkdf_extract(const bytes& key, const bytes& data) const
+{
+  const auto* type = openssl_digest_type(id);
+  auto ctx = make_typed_unique(HMAC_CTX_new());
+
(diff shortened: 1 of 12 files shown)


lib/hpke/src/digest.cpp
```diff
@@ -87,4 +87,46 @@ Digest::hmac(const bytes& key, const bytes& data) const
...
+  // https://doi.org/10.6028/NIST.SP.800-131Ar2
+  static const auto fips_min_hmac_key_len = 14;
+  auto key_size = static_cast<int>(key.size());
+  if (key_size < fips_min_hmac_key_len) {
+    HMAC_CTX_set_flags(ctx.get(), EVP_MD_CTX_FLAG_NON_FIPS_ALLOW);
+  }
+
+  // Guard against sending nullptr to HMAC_Init_ex
+  auto* key_data = key.data();
+  static const auto dummy_key = bytes{ 0 };
```
dimarusyy, line 110: Maybe better naming : default_key ?

dimarusyy (approved):

suhasHere (approved):

PR #213, opened 2021-05-23
Make TLS parsing more efficient

bifurcation:
This PR attempts to remove unnecessary copies / allocations in the TLS serialization logic.  The intent is to make parsing faster, but it's not immediately clear whether we actually achieve that objective.  The time of `make test-all` is pretty much the same on my machine, maybe even a little higher.  And since this adds some complexity for consumers, it might not be worth doing.

### DIFF
--- a/lib/tls_syntax/include/tls/tls_syntax.h
+++ b/lib/tls_syntax/include/tls/tls_syntax.h
@@ -7,10 +7,15 @@
 #include <optional>
 #include <stdexcept>
 #include <vector>
+#include <string_view>
 
 #include <tls/compat.h>
 
 namespace tls {
+
+// Abbreviations for owned and unowned buffers
+using owned_bytes = std::vector<uint8_t>;
+using input_bytes = std::basic_string_view<const uint8_t>;
 
 // For indicating no min or max in vector definitions
 const size_t none = std::numeric_limits<size_t>::max();
@@ -38,14 +43,14 @@ class ostream
 public:
   static const size_t none = std::numeric_limits<size_t>::max();
 
-  void write_raw(const std::vector<uint8_t>& bytes);
+  void write_raw(const owned_bytes& content);
 
-  std::vector<uint8_t> bytes() const { return _buffer; }
+  owned_bytes bytes() const { return _buffer; }
   size_t size() const { return _buffer.size(); }
   bool empty() const { return _buffer.empty(); }
 
 private:
-  std::vector<uint8_t> _buffer;
+  owned_bytes _buffer;
   ostream& write_uint(uint64_t value, int length);
 
   friend ostream& operator<<(ostream& out, bool data);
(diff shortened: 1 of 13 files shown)


lib/tls_syntax/include/tls/tls_syntax.h
```diff
@@ -29,24 +35,192 @@ class ReadError : public std::invalid_argument
...
+constexpr size_t size_of(const uint32_t&) { return 4; }
+
+template<>
+constexpr size_t size_of(const uint64_t&) { return 8; }
+
+// Array size
+template<typename T, size_t N>
+constexpr size_t size_of(const std::array<T, N>& data)
+{
+  auto out = size_t(0);
```
dimarusyy, line 122: Does it make sense to consider using std::accumulate ?

lib/tls_syntax/include/tls/tls_syntax.h
```diff
@@ -29,24 +35,192 @@ class ReadError : public std::invalid_argument
...
+
+// Optional size
+template<typename T>
+constexpr size_t size_of(const std::optional<T>& opt)
+{
+  if (!opt) {
+    return 1;
+  }
+
+  return 1 + size_of(opt::get(opt));
```
dimarusyy, line 137: Using ternary operator `?:` could give better performance.
Or considering adding attributes [[likely]]/[[unlikely]] would be useful here.

lib/tls_syntax/src/tls_syntax.cpp
```diff
@@ -3,22 +3,46 @@
...
 void
 ostream::write_raw(const std::vector<uint8_t>& bytes)
 {
-  // Not sure what the default argument is here
-  // NOLINTNEXTLINE(fuchsia-default-arguments)
-  _buffer.insert(_buffer.end(), bytes.begin(), bytes.end());
+  check_remaining(bytes.size());
+
+  // XXX(RLB) This is a hack around the const-ness of string_view
+  auto* data = const_cast<uint8_t*>(_buffer.data());
```
dimarusyy, line 20: Dirty hack. Sounds like some kind of monothonic buffer is needed.

lib/tls_syntax/include/tls/tls_syntax.h
bifurcation: Good idea, for this and `tls::vector::size_of`.

lib/tls_syntax/src/tls_syntax.cpp
bifurcation: What we really want here is [`std::span`](https://en.cppreference.com/w/cpp/container/span).  The alternatives are:

* Take on a C++20 dependency
* Bring in a GSL library or other similar polyfill as a dependency
* Include an implementation of the buffer type here

... none of which seem more appealing than this small hack.

PR #222, opened 2021-06-17
Re-enable hex encoding on test vectors

bifurcation:
In commit ffc31c5, I introduced a regression that caused `bytes` values to be serialized as JSON arrays of integers as opposed to hex-encoded strings.  This was due to the compiler on Linux complaining about a template for `std::vector<uint8_t>` being defined twice -- once with the JSON array serializer and once with the hex string serializer.  To avoid that problem and still get hex, this PR wraps `bytes` in a fairly transparent `HexBytes` wrapper.

### DIFF
--- a/lib/mls_vectors/include/mls_vectors/mls_vectors.h
+++ b/lib/mls_vectors/include/mls_vectors/mls_vectors.h
@@ -10,6 +10,26 @@
 
 namespace mls_vectors {
 
+// XXX(RLB) This construction is a little awkward, but otherwise, when we go to
+// serialize as JSON, some compilers have a hard time distinguishing between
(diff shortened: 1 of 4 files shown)


suhasHere (approved):

lib/mls_vectors/include/mls_vectors/mls_vectors.h
```diff
@@ -10,6 +10,26 @@
...
+struct HexBytes
+{
+  bytes data;
+
+  HexBytes() = default;
+  HexBytes(bytes data_in)
+    : data(std::move(data_in))
+  {}
+  operator const bytes&() const { return data; }
+  operator bytes&() { return data; }
```
dimarusyy, line 27: MIght be noexcept

PR #223, opened 2021-06-22
extension_data 32-bit length

franziskuskiefer:


bifurcation (approved):

bifurcation:
I swear I just changed this from 4 to 2, but you are correct that this matches the spec.

franziskuskiefer:
Well it's your change in the spec ;) https://github.com/mlswg/mls-protocol/commit/b5c460460070b2e8aed9d6620c056c913c4739d0 I think you also had it on the interop branch or somewhere else.

PR #226, opened 2021-06-28
Consistency with structs in the spec

eiqo:
Mainly cosmetic changes:
* `MLSPlaintext` was missing the `content_type` field
* Explicit structs to match the spec when serialising:
  * `MLSPlaintextTBS`, `MLSPlaintextTBM`
  * `MLSCiphertextContent`
  * `MLSPlaintextCommitContent`, `MLSPlaintextCommitAuthData` 
* Draft PR [#466](https://github.com/mlswg/mls-protocol/pull/466) was resolved in the meantime

### DIFF
--- a/include/mls/messages.h
+++ b/include/mls/messages.h
@@ -428,7 +428,32 @@ struct ApplicationData
   TLS_TRAITS(tls::vector<4>)
 };
 
-struct GroupContext;
+// struct {
+//     opaque group_id<0..255>;
+//     uint64 epoch;
+//     opaque tree_hash<0..255>;
+//     opaque confirmed_transcript_hash<0..255>;
+//     Extension extensions<0..2^16-1>;
+// } GroupContext;
+struct GroupContext
+{
+  bytes group_id;
+  epoch_t epoch;
+  bytes tree_hash;
+  bytes confirmed_transcript_hash;
+  ExtensionList extensions;
+
+  TLS_SERIALIZABLE(group_id,
+                   epoch,
+                   tree_hash,
+                   confirmed_transcript_hash,
+                   extensions)
+  TLS_TRAITS(tls::vector<1>,
+             tls::pass,
+             tls::vector<1>,
+             tls::vector<1>,
+             tls::pass)
+};
 
 enum struct ContentType : uint8_t
 {
@@ -453,6 +478,34 @@ struct Sender
   uint32_t sender{ 0 };
 
   TLS_SERIALIZABLE(sender_type, sender)
+};
+
+struct MLSPlaintextTBS
+{
+  const var::variant<GroupContext>& context;
+
+  const bytes& group_id;
+  const epoch_t& epoch;
+  const Sender& sender;
+  const bytes& authenticated_data;
+
+  const ContentType& content_type;
+  const var::variant<ApplicationData, Proposal, Commit>& content;
+
+  TLS_SERIALIZABLE(context,
+                   group_id,
+                   epoch,
+                   sender,
+                   authenticated_data,
+                   content_type,
+                   content)
+  TLS_TRAITS(tls::variant<SenderType>,
+             tls::vector<1>,
+             tls::pass,
+             tls::pass,
+             tls::vector<4>,
+             tls::pass,
+             tls::variant<ContentType>)
 };
 
 struct MLSPlaintext
@@ -584,5 +637,7 @@ TLS_VARIANT_MAP(mls::ProposalType, mls::AppAck, app_ack)
 TLS_VARIANT_MAP(mls::ContentType, mls::ApplicationData, application)
 TLS_VARIANT_MAP(mls::ContentType, mls::Proposal, proposal)
 TLS_VARIANT_MAP(mls::ContentType, mls::Commit, commit)
+
+TLS_VARIANT_MAP(mls::SenderType, mls::GroupContext, member)
 
 } // namespace tls
--- a/include/mls/messages.h
+++ b/include/mls/messages.h
@@ -508,6 +508,16 @@ struct MLSPlaintextTBS
              tls::variant<ContentType>)
 };
 
+struct MLSPlaintextTBM
+{
+  const MLSPlaintextTBS& tbs;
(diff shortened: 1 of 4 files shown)


include/mls/messages.h
```diff
@@ -455,12 +480,52 @@ struct Sender
...
+};
+
 struct MLSPlaintext
 {
   bytes group_id;
   epoch_t epoch;
   Sender sender;
   bytes authenticated_data;
+
+  ContentType content_type;
```
bifurcation, line 528: Is there a reason to make this explicit, as opposed to having it as part of the `variant`?

include/mls/messages.h
```diff
@@ -455,12 +480,52 @@ struct Sender
   TLS_SERIALIZABLE(sender_type, sender)
 };
 
+struct MLSPlaintextTBS
```
bifurcation, line 483: Can `MLSPlaintextTBS` and `MLSPlaintextTBM` move into `messages.cpp`?  It seems like they're not needed outside of signing/MAC operations.

include/mls/messages.h
```diff
@@ -527,6 +594,83 @@ struct MLSPlaintext
...
 };
 
+struct MLSPlaintextCommitContent
+{
+  const bytes& group_id;
+  const epoch_t& epoch;
+  const Sender& sender;
+  const bytes& authenticated_data;
+
+  const ContentType content_type{ ContentType::commit };
```
bifurcation, line 604: Nit: I would use `= ContentType::commit` here instead of the `{}` notation.

include/mls/messages.h
```diff
@@ -527,6 +594,83 @@ struct MLSPlaintext
...
+//     opaque signature<0..2^16-1>;
+//     optional<MAC> confirmation_tag;
+//     opaque padding<0..2^16-1>;
+// } MLSCiphertextContent;
+struct MLSCiphertextContent
+{
+  const var::variant<ApplicationData, Proposal, Commit>& content;
+  const bytes& signature;
+  const std::optional<MAC>& confirmation_tag;
+  const bytes& padding;
```
bifurcation, line 651: Since the padding is always zero and (b) we have a custom `operator<<` anyway, we could just have `const size_t padding_size` and call `encoded.resize(encoded.size() + padding_size, 0)` to add the pading.

src/key_schedule.cpp
```diff
@@ -416,8 +401,8 @@ GroupKeySource::decrypt(const bytes& sender_data_secret,
   return MLSPlaintext{ ct.group_id,
                        ct.epoch,
                        { SenderType::member, sender_data.sender },
-                       ct.content_type,
                        ct.authenticated_data,
+                       ct.content_type,
```
bifurcation, line 405: It might be good to move the construction logic from `MLSPlaintext::MLSPlaintext` to here.  That way only this file / struct would have to know about `MLSCiphertextContent`.  The content of that `MLSPlaintext` constructor seems like it would more properly go in an `operator>>` anyway.

include/mls/messages.h
```diff
@@ -455,12 +480,52 @@ struct Sender
   TLS_SERIALIZABLE(sender_type, sender)
 };
 
+struct MLSPlaintextTBS
+{
+  const var::variant<GroupContext>& context;
```
bifurcation, line 485: This is going to produce non-spec-compliant serialized objects: (a) it will always serialize the GroupContext [1], and (b) it will serialize an extra octet at the front for the SenderType [2].  Instead, we need to (a) only serialize the GroupContext when `sender_type == member` and (b) then just serialize the GroupContext.

It's not totally clear how to do this most elegantly, but the best solution that comes to my mind is:
* Extend the `tls::variant` trait to have a template parameter (say `implicit_selector`) that says whether the header is serialized
    * On encode, `if (!implicit_selector) { /* serialize selector */ }`
    * On decode, `if (implicit_selector) { /* throw; this is not possible */ }`
* Change this field type to `var::variant<GroupContext, struct{}>`
* Set the appropriate value in `MLSPlaintext::to_be_signed()` according to `sender.type`.

On a different note: I'm surprised this is passing tests!  Since we support external joins, this seems to imply that an external joiner has enough state to compute `context` for the previous group -- or the tests are broken.  If the former, we should update the spec so that external commits are signed with `context` as well.

[1] This might not be clear from the spec, but the intent of that `select` is that for cases other than `member`, nothing will be serialized, i.e., there's an unstated branch `default: struct{};`.  We should probably add that to clarify.

[2] `var::variant` serialization optimizes for the common case where you precede the variant with a selector, as for `Proposal` and `MLSPlaintext.content` (and a few places in TLS).  It requires that the selector be there, though.

include/mls/messages.h
bifurcation: Likewise for `MLSCiphertextContent` and `MLSCiphertextContentAAD` below.

bifurcation (changes requested):

include/mls/messages.h
eiqo: At the moment it is explicit in the [draft](https://messaginglayersecurity.rocks/mls-protocol/draft-ietf-mls-protocol.html#section-9), so I thought it might lead to interop issues.

include/mls/messages.h
eiqo: Yep, will do.

include/mls/messages.h
eiqo: Re [1]: Yes, this was not clear from the spec at all, though in retrospect it makes sense. Since other variants do not have such a default, it seems like this should be explicit in the spec?

include/mls/messages.h
eiqo: Correction for posterity: in mlspp, the pattern `T type; select (Obj.type)` is directly implemented in the serialisation of `var::variant`, so it does not have to be a part of the struct. What is currently not spec-compliant is any pattern `select (Obj.type)` without the preceding tag.

bifurcation:
Closing this because it has gotten stale.  We will need to do similar things in order to get to draft-13, though.

PR #227, opened 2021-07-20
More OpenMLS interop

bifurcation:
Draft PR to collect changes found in interop testing with OpenMLS.  So far:

* Added a `credential` field to the transcript test vector and send a valid signed commit
* Fixed encoding of ParentNode 
* Send an infinite lifetime in messages tests to make OpenMLS's validation logic happy

There will be more changes as we debug TreeKEM tests and protocol scripts.

There is some logging / debug code in here that should be removed before mergin.

### DIFF
--- a/lib/hpke/src/hpke.cpp
+++ b/lib/hpke/src/hpke.cpp
@@ -10,6 +10,8 @@
 #include <stdexcept>
 #include <string>
 
+#include <iostream>
+
 namespace hpke {
 
 using namespace bytes_ns::operators;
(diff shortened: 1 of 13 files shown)


bifurcation:
@eiqo @suhasHere This needs cleanup and CI passing, but some review on the merits would be helpful.

lib/hpke/src/hpke.cpp
```diff
@@ -316,6 +318,9 @@ ReceiverContext::ReceiverContext(Context&& c)
 std::optional<bytes>
 ReceiverContext::open(const bytes& aad, const bytes& ct)
 {
+  std::cout << "--- open aad=[" << to_hex(aad) << "] ct=[" << to_hex(ct) << "]"
+            << std::endl;
+
```
suhasHere, line 323: do we need it behind a flag ?

bifurcation: Ah, i'll just delete that.

PR #231, opened 2021-10-19
Signal the intended wire format for MLS messages

bifurcation:
Corresponds to https://github.com/mlswg/mls-protocol/pull/477

This ended up being a little more invasive than I expected.  The challenge is that you now need to know whether a message is going to be encrypted before you sign it, and in the case of Commits, before you generate the confirmation tag.  So either (1) every API that generates an MLSPlaintext object needs to be extended to indicate the caller's intent to encrypt, or (2) `State::encrypt` needs to re-sign objects.  (The latter doesn't work for Commit because the confirmation tag would need to be recomputed as well.)  In this PR, I have split the difference:

* On the one hand:
    * The CommitOpts struct used in `State::commit()` now has a flag for whether the commit will be encrypted
    * Application data protected with `State::protect()` is always encrypted, so `protect()` always sets the flag.
* On the other hand:
    * The `State::X_proposal()` methods generate MLSPlaintext objects with `wire_format = mls_plaintext`, then `State::encrypt()` re-signs them.

This seems like it avoids a ton of API clutter in the `State::X_proposal()` methods, while avoiding most re-signing.  There may be more elegant solutions.  But this seemed close enough on the API front to facilitate interop testing.

### DIFF
--- a/include/mls/messages.h
+++ b/include/mls/messages.h
@@ -530,11 +530,6 @@ struct MLSPlaintext
              tls::vector<2>,
              tls::pass,
              tls::pass)
-
-private:
-  // Not part of the struct, an indicator of whether this MLSPlaintext was
-  // constructed from an MLSCiphertext
-  bool decrypted;
 };
 
 // struct {
--- a/include/mls/messages.h
+++ b/include/mls/messages.h
@@ -430,6 +430,13 @@ struct ApplicationData
 
 struct GroupContext;
 
+enum struct WireFormat : uint8_t
+{
+  reserved = 0,
+  mls_plaintext = 1,
+  mls_ciphertext = 2,
+};
+
 enum struct ContentType : uint8_t
 {
   invalid = 0,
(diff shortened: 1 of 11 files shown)


include/mls/messages.h
```diff
@@ -445,6 +445,13 @@ struct ApplicationData
 
 struct GroupContext;
 
+enum struct WireFormat : uint8_t
+{
+  reserved = 0,
+  mls_plaintext = 1,
```
suhasHere, line 451: wonder what's the purpose of reserved entry ?

src/messages.cpp
```diff
@@ -236,12 +236,12 @@ MLSPlaintext::MLSPlaintext(bytes group_id_in,
                            ContentType content_type_in,
                            bytes authenticated_data_in,
                            const bytes& content_in)
-  : group_id(std::move(group_id_in))
+  : wire_format(WireFormat::mls_ciphertext)
+  , group_id(std::move(group_id_in))
   , epoch(epoch_in)
```
suhasHere, line 241: could we add a single line comment that explains the reason this version of the constructor makes wire_format to be mls_ciphertext. I feel it will be handy on quick read. thanks

suhasHere (approved):
One comment on adding a comment. LGTM. Thanks

PR #233, opened 2021-11-04
Use cascaded KDF instead of concatenation to consolidate PSKs

bifurcation:
This PR implement spec PR https://github.com/mlswg/mls-protocol/pull/490.  Note that it is only implemented at the level of the key schedule -- there is still no way for applications to actually inject PSKs.  That is a larger piece of work, reserved for later.  This is enough to get interop testing going at the key schedule level.

### DIFF
--- a/include/mls/messages.h
+++ b/include/mls/messages.h
@@ -95,6 +95,11 @@ struct PreSharedKeys
   TLS_TRAITS(tls::vector<2>)
 };
 
+struct PSKWithSecret {
+  PreSharedKeyID id;
+  bytes secret;
+};
+
 // struct {
 //     CipherSuite cipher_suite;
 //     opaque group_id<0..255>;
@@ -261,12 +266,12 @@ struct Welcome
   Welcome();
   Welcome(CipherSuite suite,
           const bytes& joiner_secret,
-          const bytes& psk_secret,
+          const std::vector<PSKWithSecret>& psks,
           const GroupInfo& group_info);
 
   void encrypt(const KeyPackage& kp, const std::optional<bytes>& path_secret);
   std::optional<int> find(const KeyPackage& kp) const;
-  GroupInfo decrypt(const bytes& joiner_secret, const bytes& psk_secret) const;
+  GroupInfo decrypt(const bytes& joiner_secret, const std::vector<PSKWithSecret>& psks) const;
 
   TLS_SERIALIZABLE(version, cipher_suite, secrets, encrypted_group_info)
   TLS_TRAITS(tls::pass, tls::pass, tls::vector<4>, tls::vector<4>)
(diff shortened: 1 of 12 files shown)


include/mls/messages.h
```diff
@@ -95,6 +95,12 @@ struct PreSharedKeys
   TLS_TRAITS(tls::vector<2>)
 };
 
+struct PSKWithSecret
+{
+  PreSharedKeyID id;
+  bytes secret;
+};
```
suhasHere, line 102: do we need serialization info for this struct ?

src/key_schedule.cpp
```diff
@@ -425,6 +425,32 @@ GroupKeySource::decrypt(const bytes& sender_data_secret,
...
+  auto psk_secret = suite.zero();
+  auto count = uint16_t(psks.size());
+  auto index = uint16_t(0);
+  for (const auto& psk : psks) {
+    auto psk_extracted = suite.hpke().kdf.extract(suite.zero(), psk.secret);
+    auto psk_label = tls::marshal(PSKLabel{ psk.id, index, count });
+    auto psk_input = suite.expand_with_label(
+      psk_extracted, "derived psk", psk_label, suite.secret_size());
+    psk_secret = suite.hpke().kdf.extract(psk_input, psk_secret);
+    index += 1;
```
suhasHere, line 449: clarification question: are we not overwriting psk_secret in each iteration ?

src/key_schedule.cpp
```diff
@@ -554,11 +581,26 @@ KeyScheduleEpoch::do_export(const std::string& label,
   return suite.expand_with_label(secret, "exporter", context_hash, size);
 }
 
+PSKWithSecret
+KeyScheduleEpoch::branch_psk(const bytes& group_id, epoch_t epoch)
+{
+  auto nonce = random_bytes(suite.secret_size());
+  return { { BranchPSK{ group_id, epoch }, nonce }, resumption_secret };
+}
```
suhasHere, line 589: where is ```BranchPSK``` defined ? Am I not seeing it ?

suhasHere (commented):
Few clarification questions , other than those, rest looks fine to me

include/mls/messages.h
bifurcation: No, it's just internal.  We're not going to send the secret around :)

src/key_schedule.cpp
bifurcation: We are, but that's ok -- we only need the last one.

src/key_schedule.cpp
bifurcation: `include/mls/messages.h`

suhasHere (approved):

PR #254, opened 2022-03-25
Introduce a Secret struct that auto-zeroizes on drop

### LINKED ISSUE #253: Apply zeroization consistently
Currently, some secrets are [zeroized in accordance with the MLS deletion schedule](https://github.com/cisco/mlspp/blob/main/src/key_schedule.cpp#L80), but others are not, notably on the main MLS key schedule.  We should zeroize more consistently.  For example, we could create a `Secret` type that would wrap `bytes` and have a destructor that would zeroize the stored data before freeing it, then consistently use that type for values that require zeroization.

bifurcation:
Fixes #253

### DIFF
--- a/include/mls/crypto.h
+++ b/include/mls/crypto.h
@@ -25,13 +25,45 @@ enum struct SignatureScheme : uint16_t
 SignatureScheme
 tls_signature_scheme(hpke::Signature::ID id);
 
-/// Cipher suites
+/// Secrets and Keys
+
+// Secret encapsulates a value that (a) should not be copied and (b) should be
+// zeroized before destruction.  The raw bytes can be accessed using the data()
+// method.  This is required for input into cryptographic operations, but should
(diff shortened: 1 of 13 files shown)


include/mls/crypto.h
```diff
@@ -25,14 +25,53 @@ enum struct SignatureScheme : uint16_t
...
 
-/// Cipher suites
+/// Secrets and Keys
+
+// Secret encapsulates a value that (a) should not be copied and (b) should be
+// zeroized before destruction.  The raw bytes can be accessed using the data()
+// method.  This is required for input into cryptographic operations, but should
+// not be used to copy out the value.
+//
+// TODO(rlb): Move method bodies to .cpp
```
suhasHere, line 35: we don't need this TODO anymore

suhasHere (approved):

bifurcation:
Replaced by #255

PR #260, opened 2022-04-26
force openssl version

suhasHere:

PR #261, opened 2022-04-26
Pin to OpenSSL 1.1

bifurcation:
The vcpkg team recently [upgraded the `openssl` package to version 3](https://github.com/microsoft/vcpkg/pull/22878).  That version is incompatible with MLSpp at the moment, so development workflows based on vcpkg (including our CI here) are broken.  To enable people to continue using OpenSSL 1.1, they created an `openssl1` package.  This PR moves to that package.


bifurcation:
Looks like this doesn't actually work, but #260 does.  Closing in favor of that.

PR #262, opened 2022-04-27
Add MLSMessage framework

bifurcation:
Corresponds to https://github.com/mlswg/mls-protocol/pull/523

This required a bit of conceptual realignment in the API, since it had been oriented around the idea that you could convert between MLSPlaintext and MLSCiphertext.  Now you have to declare your intent earlier, thus the `MessageOpts` structure.

As a side benefit, because this is a clean-slate reimplementation of MLSCiphertext, we now have the ability to add message padding.


src/messages.cpp
```diff
@@ -265,170 +265,586 @@ Sender::sender_type() const
...
+      content.content_type() == ContentType::application) {
+    throw InvalidParameterError(
+      "Application data cannot be sent as MLSPlaintext");
+  }
 
-  w << signature << confirmation_tag << bytes(padding_size, 0);
-  return w.bytes();
+  auto content_auth = MLSMessageContentAuth{ wire_format, std::move(content) };
+  auto tbs = content_auth.to_be_signed(context);
+  content_auth.auth.content_type = content_auth.content.content_type();
```
bifurcation, line 381: Do this in the `(WireFormat, MLSMessageContent)` constructor.

include/mls/messages.h
```diff
@@ -436,101 +439,171 @@ struct Sender
...
-            const GroupContext& context,
-            const SignaturePrivateKey& priv);
-  bool verify(const CipherSuite& suite,
-              const GroupContext& context,
-              const SignaturePublicKey& pub) const;
+struct MLSMessageContentAuth
+{
+  WireFormat wire_format;
+  MLSMessageContent content;
+  MLSMessageAuth auth;
```
bifurcation, line 480: Make these private and expose accessors to any needed fields.

src/messages.cpp
```diff
@@ -265,170 +265,586 @@ Sender::sender_type() const
...
+  : wire_format(wire_format_in)
+  , content(std::move(content_in))
+  , auth(std::move(auth_in))
+{}
+
+bytes
+MLSMessageContentAuth::to_be_signed(
+  const std::optional<GroupContext>& context) const
+{
+  return tls::marshal(MLSMessageContentTBS{
```
bifurcation, line 486: Move MLSMessageContentTBS down here.

src/messages.cpp
```diff
@@ -265,170 +265,586 @@ Sender::sender_type() const
...
+    throw InvalidParameterError("Wire format mismatch (not mls_plaintext)");
+  }
+}
+
+bytes
+MLSPlaintext::membership_mac(CipherSuite suite,
+                             const bytes& membership_key,
+                             const std::optional<GroupContext>& context) const
+{
+  auto tbm = tls::marshal(MLSMessageContentTBM{
```
bifurcation, line 555: Move MLSMessageContentTBM down here.

src/messages.cpp
```diff
@@ -265,170 +265,586 @@ Sender::sender_type() const
...
+tls::ostream&
+operator<<(tls::ostream& str, const MLSPlaintext& obj)
+{
+  str << obj.content << obj.auth;
+
+  if (obj.content.sender.sender_type() == SenderType::member) {
+    str << opt::get(obj.membership_tag);
+  }
+
+  return str;
```
bifurcation, line 572: Refactor as `switch (obj.content.sender_type())`

src/messages.cpp
```diff
@@ -265,170 +265,586 @@ Sender::sender_type() const
...
+operator>>(tls::istream& str, MLSPlaintext& obj)
+{
+  str >> obj.content;
+
+  obj.auth.content_type = obj.content.content_type();
+  str >> obj.auth;
+
+  if (obj.content.sender.sender_type() == SenderType::member) {
+    obj.membership_tag.emplace();
+    str >> opt::get(obj.membership_tag);
```
bifurcation, line 585: Refactor as `switch (obj.content.sender_type())`

src/messages.cpp
```diff
@@ -265,170 +265,586 @@ Sender::sender_type() const
...
+//         case proposal:
+//           Proposal proposal;
+//
+//         case commit:
+//           Commit commit;
+//     }
+//
+//     MLSMessageAuth auth;
+//     opaque padding<V>;
+// } MLSCiphertextContent;
```
bifurcation, line 605:
```suggestion
```

src/messages.cpp
```diff
@@ -265,170 +265,586 @@ Sender::sender_type() const
...
+      content.content.emplace<Proposal>();
+      break;
+
+    case ContentType::commit:
+      content.content.emplace<Commit>();
+      break;
+
+    default:
+      throw InvalidParameterError("Unknown content type");
+  }
```
bifurcation, line 639: Do this switch in `unprotect`.  Then the `visit()` here will have the intended effect.

include/mls/state.h
```diff
@@ -129,14 +141,16 @@ class State
...
+  // next_state option into CommitOpts to pull this into State?
+  MLSMessageContentAuth unprotect(const MLSMessage& msg);
 
   ///
   /// Application encryption and decryption
   ///
-  MLSCiphertext protect(const bytes& pt);
-  bytes unprotect(const MLSCiphertext& ct);
+  MLSMessage protect_app(const bytes& pt, const MessageOpts& msg_opts);
+  bytes unprotect_app(const MLSMessage& ct);
```
bifurcation, line 153: This naming is unfortunate.  Maybe we could transfer the ugliness to the private API and have the public API be nicer.

bifurcation (commented):
Covers all by `state.cpp.

src/state.cpp
```diff
@@ -205,23 +207,96 @@ State::external_join(const bytes& leaf_secret,
...
+
+    default:
+      throw InvalidParameterError("Malformed MLSMessageContentAuth");
+  }
+}
+
+MLSMessageContentAuth
+State::unprotect(const MLSMessage& msg)
+{
+
```
bifurcation, line 274:
```suggestion
```

src/state.cpp
```diff
@@ -205,23 +207,96 @@ State::external_join(const bytes& leaf_secret,
...
+      auto maybe_content_auth =
+        ct.unprotect(_suite, _tree, _keys, _key_schedule.sender_data_secret);
+      if (!maybe_content_auth) {
+        throw ProtocolError("MLSCiphertext decryption failure");
+      }
+      return opt::get(maybe_content_auth);
+    },
+
+    [](const auto& /* unused */) -> MLSMessageContentAuth {
+      throw ProtocolError("Invalid message type");
```
bifurcation, line 295:
```suggestion
      throw ProtocolError("Invalid wire format");
```

src/state.cpp
```diff
@@ -436,14 +516,23 @@ State::commit(const bytes& leaf_secret,
...
   // Create the Commit message and advance the transcripts / key schedule
-  auto encrypt_handshake = opts && opt::get(opts).encrypt_handshake;
-  auto pt = next.ratchet_and_sign(sender,
-                                  commit,
-                                  commit_secret,
-                                  { /* no PSKs */ },
-                                  force_init_secret,
-                                  encrypt_handshake,
-                                  group_context());
+  // TODO(rlb) move this back into ratchet_and_sign?
```
bifurcation, line 519:
```suggestion
```

src/state.cpp
```diff
@@ -484,63 +573,56 @@ State::group_context() const
...
+  return handle(unprotect(msg));
 }
 
 std::optional<State>
-State::handle(const MLSPlaintext& pt)
+State::handle(const MLSMessageContentAuth& content_auth)
 {
-  // Pre-validate the MLSPlaintext
-  if (pt.group_id != _group_id) {
+  // Verify the signature on the message and basic parameters
```
bifurcation, line 590:
```suggestion
  // Verify the signature on the message
```

src/state.cpp
```diff
@@ -773,16 +853,17 @@ State::extensions_supported(const ExtensionList& exts) const
...
 void
-State::cache_proposal(const MLSPlaintext& pt)
+State::cache_proposal(MLSMessageContentAuth content_auth)
 {
   auto sender_location = std::optional<LeafIndex>();
-  if (pt.sender.sender_type() == SenderType::member) {
-    sender_location = _tree.find(var::get<LeafNodeRef>(pt.sender.sender));
+  if (content_auth.content.sender.sender_type() == SenderType::member) {
+    sender_location =
+      _tree.find(var::get<LeafNodeRef>(content_auth.content.sender.sender));
```
bifurcation, line 861:
```suggestion
    const auto& sender = content_auth.content.sender.sender;
    sender_location = _tree.find(var::get<LeafNodeRef>(sender));
```

src/state.cpp
```diff
@@ -905,31 +986,35 @@ State::apply(const std::vector<CachedProposal>& proposals)
 /// Message protection
 ///
 
-MLSCiphertext
-State::protect(const bytes& pt)
+MLSMessage
+State::protect_app(const bytes& pt, const MessageOpts& msg_opts)
```
bifurcation, line 990: Change signature to take `(authenticated_data, padding_size)` instead of `MessageOpts`, since `encrypt` isn't relevant.

src/state.cpp
```diff
@@ -905,31 +986,35 @@ State::apply(const std::vector<CachedProposal>& proposals)
...
+  if (!msg_opts.encrypt) {
+    throw InvalidParameterError("Application data cannot be sent as plaintext");
+  }
+
+  return protect_full(ApplicationData{ pt }, msg_opts);
 }
 
 bytes
-State::unprotect(const MLSCiphertext& ct)
+State::unprotect_app(const MLSMessage& ct)
```
bifurcation, line 1000: Should this return `authenticated_data` somehow?

src/state.cpp
```diff
@@ -905,31 +986,35 @@ State::apply(const std::vector<CachedProposal>& proposals)
...
 {
-  MLSPlaintext pt = decrypt(ct);
+  auto content_auth = unprotect(ct);
 
-  if (!verify(pt)) {
-    throw ProtocolError("Invalid message signature");
+  if (!verify(content_auth)) {
+    // TODO(rlb): Verify that sender is internal (!)
```
bifurcation, line 1005:
```suggestion
```

include/mls/state.h
```diff
@@ -129,14 +141,16 @@ class State
   ///
   /// General encryption and decryption
   ///
-  MLSCiphertext encrypt(const MLSPlaintext& pt);
-  MLSPlaintext decrypt(const MLSCiphertext& ct);
+  // XXX(RLB) It would be good to hide this from upper layers.  However, Session
```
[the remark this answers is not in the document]
bifurcation: As a side note: I briefly considered suggesting that Session could just read the unauthenticated fields, but of course it's a bad idea to operate on unauthenticated data.  That just makes the case extra strong for handling this caching inside of `State::handle`, after unprotecting and verifying the message.

Putting the picture together, I think it looks like:

* Add an argument `std::optional<State> cached_state` to `::handle()`
* Inside `State::handle`: `if (content_type == commit && cached_state) { return cached_state; }`
* In Session, provide cached state iff processing a commit from ourselves

include/mls/key_schedule.h
```diff
@@ -84,6 +82,7 @@ struct GroupKeySource
   std::map<Key, HashRatchet> chains;
 
   HashRatchet& chain(RatchetType type, LeafIndex sender);
+  HashRatchet& chain(ContentType type, LeafIndex sender);
```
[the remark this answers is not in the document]
bifurcation: Filed #263

include/mls/messages.h
```diff
@@ -397,6 +397,9 @@ enum struct WireFormat : uint8_t
   reserved = 0,
   mls_plaintext = 1,
   mls_ciphertext = 2,
+  mls_welcome = 3,
+  mls_group_info = 4,
+  mls_key_package = 5,
```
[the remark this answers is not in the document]
bifurcation: Filed #264

include/mls/state.h
```diff
@@ -91,20 +98,25 @@ class State
...
-    const std::optional<CommitOpts>& opts) const;
+    const std::optional<CommitOpts>& opts,
+    const MessageOpts& msg_opts);
 
   ///
   /// Generic handshake message handler
   ///
-  std::optional<State> handle(const MLSPlaintext& pt);
+  std::optional<State> handle(const MLSMessage& msg);
+  std::optional<State> handle(const MLSMessageContentAuth& content_auth);
```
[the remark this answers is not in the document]
bifurcation: We need this as long as Session is doing cached state detection by calling `unprotect()`.

src/messages.cpp
bifurcation: Actually, this doesn't work well, since the logic to set `auth.content_type` would have to get replicated.

include/mls/state.h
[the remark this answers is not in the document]
bifurcation: Went ahead and fixed this.

src/messages.cpp
```diff
@@ -265,170 +265,580 @@ Sender::sender_type() const
...
+    epoch,
+    content_type,
+  });
+
+  auto sender_data_pt = suite.hpke().aead.open(sender_data_keys.key,
+                                               sender_data_keys.nonce,
+                                               sender_data_aad,
+                                               encrypted_sender_data);
+  if (!sender_data_pt) {
+    return {};
```
suhasHere, line 748: should the return be std::nullopt instead ?

suhasHere (approved):
Looks good overall , have few nits.

include/mls/messages.h
```diff
@@ -436,101 +439,183 @@ struct Sender
...
+  epoch_t epoch() const;
+  WireFormat wire_format() const;
+
+  MLSMessage() = default;
+  MLSMessage(MLSPlaintext mls_plaintext);
+  MLSMessage(MLSCiphertext mls_ciphertext);
+  MLSMessage(Welcome welcome);
+  MLSMessage(GroupInfo group_info);
+  MLSMessage(KeyPackage key_package);
+
```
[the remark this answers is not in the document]
bifurcation: We're going to `std::move` these into storage.  Providing them by value like this means that if a caller provides an rvalue, there's no copying (the rvalue reference constructor for the argument will be invoked).  Something like:

```
auto mls_message = MLSMessage{ Welcome{ ... } };
// or 
auto welcome = Welcome{ ... };
auto mls_message = MLSMessage{ std::move(welcome) };
```

If this were `const Welcome&` (etc), then there would always be a copy.

include/mls/messages.h
```diff
@@ -436,101 +439,183 @@ struct Sender
...
+  MLSMessageAuth auth;
 
-  bytes membership_tag_input(const GroupContext& context) const;
-  bool verify_membership_tag(const bytes& tag) const;
+  MLSMessageContentAuth() = default;
 
-  bytes marshal_content(size_t padding_size) const;
+  static MLSMessageContentAuth sign(WireFormat wire_format,
+                                    MLSMessageContent content,
+                                    CipherSuite suite,
```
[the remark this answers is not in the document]
bifurcation: Same reasoning as with MLSMessage; we `std::move` content in here, so this allows for a no-copy call.

include/mls/state.h
```diff
@@ -181,25 +187,26 @@ class State
...
 
-  // Form a commit that can be either internal or external
-  std::tuple<MLSPlaintext, Welcome, State> commit(
+  std::tuple<MLSMessage, Welcome, State> commit(
     const bytes& leaf_secret,
     const std::optional<CommitOpts>& opts,
+    const MessageOpts& msg_opts,
     const std::optional<KeyPackage>& joiner_key_package,
-    const std::optional<HPKEPublicKey>& external_pub) const;
+    const std::optional<HPKEPublicKey>& external_pub);
```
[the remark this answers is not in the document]
bifurcation: Thinking here is to keep it together with CommitOpts.

src/messages.cpp
bifurcation: Sure.  It's equivalent, since `std::optional<T>` default-initializes to `std::nullopt`.  But it's clearer to be explicit.

PR #283, opened 2022-07-27
Support negotiation of credential types

bifurcation:
Corresponds to https://github.com/mlswg/mls-protocol/pull/631

### DIFF
--- a/src/core_types.cpp
+++ b/src/core_types.cpp
@@ -43,6 +43,11 @@ const std::array<CipherSuite::ID, 6> all_supported_ciphersuites = {
   CipherSuite::ID::X448_CHACHA20POLY1305_SHA512_Ed448,
 };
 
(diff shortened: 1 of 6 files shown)


src/core_types.cpp
```diff
@@ -95,6 +101,13 @@ Capabilities::proposals_supported(
     });
 }
 
+bool
+Capabilities::credential_supported(const Credential& credential) const
+{
+  const auto cred_type = credential.type();
+  return std::count(credentials.begin(), credentials.end(), cred_type) > 0;
+}
```
suhasHere, line 109: Nit .. may be simplify it to one line ?

suhasHere (approved):

src/core_types.cpp
bifurcation: `clang-format` was wrapping it, so i thought it would be a little more elegant this way.

PR #286, opened 2022-07-30
Make tree math interface more object-oriented

bifurcation:
Depends on #285


src/tree_math.cpp
```diff
@@ -57,9 +57,25 @@ LeafIndex::LeafIndex(NodeIndex x)
...
+  uint8_t k = 0;
+  while (ln != rn) {
+    ln.val = ln.val >> 1U;
+    rn.val = rn.val >> 1U;
+    k += 1;
+  }
+
+  uint32_t prefix = ln.val << k;
+  uint32_t stop = (1U << uint8_t(k - 1));
+  return NodeIndex{ prefix + (stop - 1) };
```
suhasHere, line 78: wondering if a basic test would be useful ?

suhasHere (approved):

src/tree_math.cpp
bifurcation: Good idea, added ancestor to test vectors.

PR #287, opened 2022-07-30
Suppress generation of redundant nodes

bifurcation:
Corresponds to https://github.com/mlswg/mls-protocol/pull/587

Depends on #285 and #286


src/key_schedule.cpp
```diff
@@ -148,23 +148,23 @@ SecretTree::get(LeafIndex sender)
...
-    auto& secret = secrets[curr_node.val];
-    secrets[left.val] =
-      derive_tree_secret(suite, secret, "tree", left, 0, secret_size);
-    secrets[right.val] =
-      derive_tree_secret(suite, secret, "tree", right, 0, secret_size);
+    auto curr_node = dirpath.at(curr);
+    auto left = curr_node.left();
+    auto right = curr_node.right();
+
+    auto& secret = secrets.at(curr_node);
```
suhasHere, line 155: do we want to ensure against this lookup failing ?

suhasHere (approved):
just one comment .. overall the flow  seems to match the spec update in general

src/key_schedule.cpp
bifurcation: `.at()` throws if the key doesn't exist, so it performs the check for us.

suhasHere: ah ! that's right .. thanks

PR #289, opened 2022-08-01
Use tree hash as input to parent hash

bifurcation:
Corresponds to https://github.com/mlswg/mls-protocol/pull/527

Depends on #284 #285 #286 #287


src/treekem.cpp
```diff
@@ -869,6 +876,81 @@ TreeKEMPublicKey::parent_hashes(
...
+
+  // If there are no local changes, then we can use the cached tree hash
+  if (!have_local_changes) {
+    return hashes.at(index);
+  }
+
+  // If this method has been called before with the same number of excluded
+  // leaves (which implies the same set), then use the cached value.
+  auto cache_key = std::make_pair(index, except.size());
+  if (auto it = cache.find(cache_key); it != cache.end()) {
```
suhasHere, line 902: just curious, is it possible that same sized excluded set can happen but with different operations 
on the tree between the calls here ?
I want to make sure the excluded set size check is a safe assumption

bifurcation: Good question.  That came from some code @TWal added as we were prototyping this.  I believe the argument goes as follows:

* The possible `except` sets here are the intersection of the `unmerged_leaves` of ancestor nodes with the current subtree
* Because nodes always get reset along a direct path, if a leaf is unmerged at a node `P`, then it unmerged at all the non-blank nodes between the leaf and `P`.
* As a result, the `except` sets relative to ancestors are all subsets of one another.  If there are ancestors `P1`, ..., `Pk` from the current node to the root with corresponding `except` sets `E1, ..., Ek`, then `E1 >= E2 >= ... >= Ek`.
* That means that among the possible `exclude` sets, there's only one with a given size.

TWal: Yes, this is the argument indeed!

However, when working with parent-hash security proofs, I realized there are two kind of invariants:
- the ones true for group creation (base case), and group modification (induction case)
- the ones true for group creation AND accepting a Welcome message (base case), and group modifications (induction case)

The correctness of this implementation relies on an invariant of the first kind, which is not checked when accepting a Welcome message.
This implementation won't accept trees that are rejected by the RFC, however it might reject trees that are accepted by the RFC (in this case it means that someone sent you a bogus tree, so things are weird anyway…)

Something we could do is to add to the Welcome check all the invariants of MLS, so that if the Welcome check succeeds, there exists a sequence of operations that starting from a group creation, that gives the tree you are processing. Hence, the first kind of invariants and the second kind of invariants would be the same.

src/treekem.cpp
```diff
@@ -493,22 +493,29 @@ TreeKEMPublicKey::has_parent_hash(NodeIndex child, const bytes& target_ph) const
...
+
   auto width = NodeCount(size);
-  for (auto i = NodeIndex{ 1 }; i.val < width.val; i.val += 2) {
-    if (node_at(i).blank()) {
-      continue;
-    }
+  auto height = NodeIndex::root(size).level();
+  for (auto level = uint32_t(1); level <= height; level++) {
+    auto stride = uint32_t(2) << level;
+    auto start = NodeIndex{ stride - 1 };
```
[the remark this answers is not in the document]
bifurcation: Ah yes, that looks right, thanks.  I might do it as `stride >> 1` to make the relationship clear.

include/mls/treekem.h
```diff
@@ -179,6 +179,14 @@ struct TreeKEMPublicKey
     const FilteredDirectPath& fdp,
     const std::vector<UpdatePathNode>& path_nodes) const;
 
+  using TreeHashCache = std::map<std::tuple<NodeIndex, size_t>, bytes>;
```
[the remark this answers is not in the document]
bifurcation: Just to confirm I understand, this works because the parent hash algorithm traverses the tree in "one direction", in the sense of the `E1 >= E2 >= ... >= Ek` chain above?  So the cache captures the cases where `Ei = ... = Ej` for some span.  Want to capture any dependency between the caching strategy and the traversal to avoid breaking one by changing the other.

Otherwise, this doesn't seem too intimdating, will give it a try.

suhasHere (approved):

PR #291, opened 2022-08-09
Keep unmerged leaves in order

bifurcation:
Corresponds to https://github.com/mlswg/mls-protocol/pull/714


suhasHere (approved):

TWal:
If I'm not mistaken, you also need to check unmerged leaves are sorted when processing a Welcome message?

Issue #259, opened 2022-04-26
Support OpenSSL 3.0


bifurcation:
Currently, the `hpke` crypto library abstractions call several functions that have been deprecated in OpenSSL 3.0.  Since the GitHub Actions runners appear to have upgraded to OpenSSL 3.0, this means that CI is broken until we support it.

The functions we rely on fall in basically three categories:

* `EC_KEY_*`
  * `EC_KEY_free`
  * `EC_KEY_get0_group`
  * `EC_KEY_get0_private_key`
  * `EC_KEY_new_by_curve_name`
  * `EC_KEY_set_private_key`
  * `EC_KEY_set_public_key`
  * `EVP_PKEY_assign`
  * `EVP_PKEY_get0_EC_KEY`
* Serialization / deserialization
  * `i2o_ECPublicKey`
  * `o2i_ECPublicKey`
* HMAC
  * `HMAC_CTX_free`
  * `HMAC_CTX_new`
  * `HMAC_CTX_set_flags`
  * `HMAC_Final`
  * `HMAC_Init_ex`
  * `HMAC_Update`

We should not switch over entirely to 3.0, since it's pretty new.  So we will need CMake tooling to detect what version we have, set a `#define` configuration, and then have both 1.1 and 3.0 implementations.

Useful references:
* https://www.openssl.org/docs/man3.0/man7/migration_guide.html
* https://cmake.org/cmake/help/latest/module/FindOpenSSL.html

ctjhai:
Hi @bifurcation, I have an implementation that works as suggested above, i.e. let CMake detects the version of OpenSSL available in the system. However, there is a minor issue in that it may not be possible to set the flag `EVP_MD_CTX_FLAG_NON_FIPS_ALLOW` in FIPS mode. I think OpenSSL 3.0 doesn't support this flag anymore. It may be possible to set the flag as follows (although I am not sure if it does anything):
```c
/* Error checking is ignored */
EVP_PKEY *key = EVP_PKEY_new_mac_key(EVP_PKEY_HMAC, NULL, key_data, key_size);
EVP_MD_CTX *mctx = EVP_MD_CTX_new();
EVP_MD *md = EVP_sha256();
EVP_PKEY_CTX *pctx = NULL;
EVP_MD_CTX_set_flags(mctx, EVP_MD_CTX_FLAG_NON_FIPS_ALLOW); /* <-- Set the flag here */
EVP_DigestSignInit(mctx, &pctx, md, NULL, key);
[...]

PR #295, opened 2023-01-19
Updated to support OpenSSL 3

ctjhai:
This PR aims to address issue https://github.com/cisco/mlspp/issues/259.


ctjhai:
Hi @bifurcation, as a follow up of our discussion in https://github.com/cisco/mlspp/issues/259, I've been looking at OpenSSL 3.x source code, and found that the flag `EVP_MD_CTX_FLAG_NON_FIPS_ALLOW` is not used anymore, see:
https://github.com/openssl/openssl/blob/3147785eb23bb27080a0b7accbbff46ac471e86c/include/openssl/evp.h#L208-L209

The discussion around this can be found here: https://github.com/openssl/openssl/pull/17484.

I also found that in some OpenSSL versions (1.1.1) that I've tested, the flag `EVP_MD_CTX_FLAG_NON_FIPS_ALLOW` is only used to load non-FIPS implementation, e.g. loading MD5 implementation in FIPS mode; it doesn't really validate the constraint on the key size. I guess some custom implementation of the FIPS object module may be stricter, so perhaps this flag has been used to override this behaviour?

Anyway, with OpenSSL 3.x, perhaps we need to ensure that both FIPS and default providers are loaded, so that we can fallback to the default implementation to allow a more relaxed key size. So there is no need to force the use OpenSSL 1.1.1 for the FIPS mode. Having said that, just like version 1.1.1, it looks like OpenSSL 3.x is happy with HMAC key size < 112 bits even in FIPS mode. I tried 0-length HMAC key size with [OpenSSL 3.0 FIPS provider](https://csrc.nist.gov/CSRC/media/projects/cryptographic-module-validation-program/documents/security-policies/140sp4282.pdf), there is no issue; I arrive at the expected HMAC digest. 

There are some test cases that fail in FIPS mode though, but I think this is related to using the likes of X25519 and X448 for key-agreement or Ed25519 in the X.509. I'll update the test cases as part of this pull-request.

bifurcation (changes requested):
This is a great start, thanks @ctjhai.  A few suggestions for refactoring, and it looks like you have some CI issues to clear, but overall this seems like the right direction.

Given the depth of the changes in `group.cpp` and `digest.cpp`, I wonder if it would be simpler just to have two versions of the files (one with the v1 implementation and one with v3), and switch between them in CMake.  One of my concerns here is how easy it's going to be to revert this change once we're more in an all-v3 world, and that would certainly make that reversion simple.

It would also be good to add a CI run here that builds with OpenSSL v3.  Might need an alternate `vcpkg.json` that gets swapped in at the right moment.

lib/hpke/src/certificate.cpp
```diff
@@ -199,8 +199,8 @@ struct Certificate::ParsedCertificate
     , sub_alt_names(parse_san(x509.get()))
     , is_ca(X509_check_ca(x509.get()) != 0)
     , hash(compute_digest(x509.get()))
-    , not_before(asn1_time_to_chrono(X509_get_notBefore(x509.get())))
-    , not_after(asn1_time_to_chrono(X509_get_notAfter(x509.get())))
+    , not_before(asn1_time_to_chrono(X509_getm_notBefore(x509.get())))
+    , not_after(asn1_time_to_chrono(X509_getm_notAfter(x509.get())))
```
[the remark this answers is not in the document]
ctjhai: My bad, this should be `X509_get0_XXX`.

lib/hpke/src/digest.cpp
```diff
@@ -102,9 +139,21 @@ Digest::hmac_for_hkdf_extract(const bytes& key, const bytes& data) const
...
   static const auto fips_min_hmac_key_len = 14;
   auto key_size = static_cast<int>(key.size());
+#if defined(WITH_OPENSSL3)
+  if (EVP_default_properties_is_fips_enabled(nullptr) &&
+      key_size < fips_min_hmac_key_len) {
+    // Currently OpenSSL's EVP_PKEY_new_mac_key API cannot deal with an empty
+    // key, see: https://github.com/openssl/openssl/issues/13089 As such, we
+    // have to use EVP_MAC API to do the HMAC operations. I am not sure if it's
+    // possible to do the following (or the equivalent) with EVP_MAC APIs:
+    //   EVP_MD_CTX_set_flags(context, EVP_MD_CTX_FLAG_NON_FIPS_ALLOW)
```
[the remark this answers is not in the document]
ctjhai: Since the flag `EVP_MD_CTX_FLAG_NON_FIPS_ALLOW` has no effect from OpenSSL 3.0, I don't think we need to wait for this bug to be fixed. We should just use `EVP_MAC` APIs.

As far as I understand, OpenSSL doesn't seem to validate the key size of a HMAC in FIPS mode; it will validate whether an algorithm is FIPS approved, for example it will complain if we specify MD5 algorithm in FIPS mode. From what I've experimented so far, even if with just FIPS provider loaded, OpenSSL 3 doesn't complain if the HMAC key size is less than 112 bits, an expected digest value is always returned. Therefore, I think there is no need to check for the key size violation here. If in the future, OpenSSL becomes stricter in FIPS mode, we will need to handle this issue then. One way to handle this is to ensure both FIPS and default providers are loaded so that the shorter key can be provided by the fallback default provider.

lib/hpke/src/group.cpp
```diff
@@ -162,18 +251,22 @@ struct ECKeyGroup : public EVPGroup
 
     auto dkp_prk = kdf.labeled_extract(suite_id, {}, label_dkp_prk, ikm);
 
+#if defined(WITH_OPENSSL3)
+    EC_GROUP* group =
+      EC_GROUP_new_by_curve_name_ex(nullptr, nullptr, curve_nid);
+    auto group_ptr = make_typed_unique(group);
```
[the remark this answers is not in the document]
ctjhai: I am not sure if we can remove the `#if`/`#else` definition here. If I understand it correctly, the `EC_GROUP` pointer returned by `EC_KEY_get0_group` (i.e. OpenSSL 1.1.1) doesn't need to be freed, whereas the one returned by `EC_GROUP_new_by_curve_name_ex` (i.e. OpenSSL 3.0) needs to be freed, hence it is wrapped with `make_typed_unique`.

ctjhai:
> Given the depth of the changes in `group.cpp` and `digest.cpp`, I wonder if it would be simpler just to have two versions of the files (one with the v1 implementation and one with v3), and switch between them in CMake. One of my concerns here is how easy it's going to be to revert this change once we're more in an all-v3 world, and that would certainly make that reversion simple.

The changes in `digest.cpp` are pretty small compared to `group.cpp`, so I am not entirely sure about creating a separately file for this. Additionally, another advantage of not creating two files is we can easily access the trails of changes of the files (from `git blame` perspective). In terms of reversion, we can use tools like [unifdef](https://dotat.at/prog/unifdef/) to pick the code for OpenSSL 3 only.

Having said that, if you insist, I am happy to use two separate files for these. But we still need to cater for the `WITH_OPENSSL3` flag in `openssl_common.cpp`. So would you also like to have two files for `openssl_common.cpp` to maintain consistency?

lib/hpke/src/group.cpp
```diff
@@ -152,6 +156,91 @@ struct ECKeyGroup : public EVPGroup
   {
   }
 
+#if defined(WITH_OPENSSL3)
+  int EVP_PKEY_set_keys(EVP_PKEY* key, const bytes& sk, const bytes& pk) const
```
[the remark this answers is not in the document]
ctjhai: This is now refactored into two functions:
- `typed_unique_ptr<EVP_PKEY> keypair_evp_key(const typed_unique_ptr<BIGNUM>& priv) const`, and
- `typed_unique_ptr<EVP_PKEY> public_evp_key(const bytes& pub) const`

I renamed the first one to use `keypair` instead of `private` to denote that the returned `EVP_PKEY` object to always contain both private and public keys.

.github/workflows/compat_openssl3.yml
```diff
@@ -0,0 +1,43 @@
+name: Build for older MacOS (OpenSSL 3)
```
bifurcation, line 1: We don't need this extra build.  The MacOS compat issues are orthogonal to OpenSSL 3.

CMakeLists.txt
```diff
@@ -70,14 +70,27 @@ endif()
 ### Dependencies
 ###
 
+# External libraries
+find_package(OpenSSL REQUIRED)
+if ( OPENSSL_FOUND )
+    if (${OPENSSL_VERSION} VERSION_GREATER 1.1.1)
```
bifurcation, line 76:
```suggestion
    if (${OPENSSL_VERSION} VERSION_GREATER_EQUAL 3)
```

bifurcation: (This avoids spurious errors if the 1.X line evolves)

lib/hpke/src/digest.cpp
```diff
@@ -91,20 +114,40 @@ Digest::hmac(const bytes& key, const bytes& data) const
...
+#if defined(WITH_OPENSSL3)
+  auto mac =
+    make_typed_unique(EVP_MAC_fetch(nullptr, OSSL_MAC_NAME_HMAC, nullptr));
+  auto ctx = make_typed_unique(EVP_MAC_CTX_new(mac.get()));
+  auto digest_name = openssl_digest_name(id);
+  std::array<OSSL_PARAM, 2> params = {
+    OSSL_PARAM_construct_utf8_string(
+      OSSL_ALG_PARAM_DIGEST, digest_name.data(), 0),
+    OSSL_PARAM_construct_end()
+  };
```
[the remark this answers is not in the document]
bifurcation: (`const auto` if possible)

lib/hpke/test/common.cpp
```diff
@@ -5,6 +5,35 @@
...
+static OSSL_PROVIDER* fips_prov = nullptr;
+
+static int
+FIPS_mode()
+{
+  if (OSSL_PROVIDER_available(nullptr, "fips") == 1) {
+    return 1;
+  }
+  return EVP_default_properties_is_fips_enabled(nullptr);
+}
```
bifurcation, line 21: Rather than emulate the old OpenSSL FIPS API, I would pull the contents of `fips.cpp/h` into this file and its header, and make variants of `fips()` and `enable_fips()` for the two versions.  Where in the latter case, I mean either make a new function `enable_fips()` that varies between the versions, or just branch internally to `ensure_fips_if_required()`.

lib/mls_vectors/test/fips.h
```diff
@@ -0,0 +1,7 @@
+#include <mls/crypto.h>
+
+bool
+fips();
+
+bool
+is_fips_approved(mls::CipherSuite::ID id);
```
bifurcation, line 7: These methods seem duplicative of `fips` and `fips_disable`?

test/fips.h
```diff
@@ -0,0 +1,7 @@
+#include <mls/crypto.h>
+
+bool
+fips();
+
+bool
+is_fips_approved(mls::CipherSuite::ID id);
```
bifurcation, line 7: If we need these methods in `test`, let's just make them a public API of the `hpke` library.  We shouldn't be directly including OpenSSL headers outside of HPKE.

bifurcation (changes requested):
Thanks @ctjhai, this is on the right track.  A couple of minor things, then this is good to go.

lib/hpke/test/common.cpp
ctjhai: Not a problem

lib/mls_vectors/test/fips.h
ctjhai: This method is no longer needed, so removed in the current commit.

test/fips.h
ctjhai: This method is no longer needed. Initially, I needed it when I tested my changes on an Oracle Linux 8 with FIPS enabled. The OS came with OpenSSL 1.1.1k FIPS and it had a smaller set of FIPS approved algorithms compared to [OpenSSL 3.0 FIPS](https://csrc.nist.gov/CSRC/media/projects/cryptographic-module-validation-program/documents/security-policies/140sp4282.pdf). So I had to use the above method to skip certain algorithm choices. Given that we are moving towards the later version of OpenSSL, I don't see there is need to skip any test so I think we can remove this method.

bifurcation (approved):
Thanks for all the work on this @ctjhai !

bifurcation:
@ctjhai - Some CI runs are failing due to deficiencies in the macOS GitHub Actions runners.  These were fixed in #314, but your PR is older than that.  If you could please do a quick `git checkout main && git pull && git checkout openssl3 && git merge main`, that should turn the CI green and I'll merge.

PR #297, opened 2023-02-03
Start bringing interop runner up to current spec

bifurcation:
* Update the test runner to have explicit functions to generate and verify test vector files
* Update the tree math and encryption test vector processing to match the current definition
* Remove the logging facility that only ever really got half-way implemented
* Fix bugs discovered so far in interop testing:
    * LeafCount::full produced inaccurate results for one-member tree (2 vs 1)
    * LeafIndex TLS-encoding as NodeIndex
    * `mls10 -> MLS 1.0` in ExpandWithLabel
    * No `TreeContext{ node, generation }` in DeriveTreeSecret
    * Derivation in secret tree by left/right, not by node index
    * MLSPlaintext/MLSCiphertext protect/unprotect unnecessarily depended on ratchet tree


suhasHere (approved):
Changes in the library looks fine to me. The one in vectors seemed fine in general but I wasn't able to look carefully .. I can do it async.

PR #302, opened 2023-02-07
Update struct names to match draft-17

bifurcation:
Corresponds to https://github.com/mlswg/mls-protocol/pull/839


suhasHere (approved):
Lot of changes , but they all seem to fall under the name changes and not logical changes, unless I am missing something
LGTM

PR #305, opened 2023-02-09
Make test vectors deterministic

bifurcation:
This PR makes the values included in test vectors deterministic, to make it easier to see what has changed from one test vector generation to the next.  Basically, for each test vector, we create a random seed from the type of the test vector, then use `ExpandWithLabel` to produce random values by label.  Most of the diff is thus changing calls to `random_bytes` to calls to the PRG.

While I was here, I also changed the remaining static `create()` methods to constructors.


suhasHere (approved):
Looks fine to me

PR #314, opened 2023-02-20
Install pkg-config on macOS CI runners

bifurcation:
Apparently this used to be present, but has been removed.


bifurcation:
Merging because no code is touched and this is blocking other PRs.

PR #315, opened 2023-02-20
Separate TreeKEM update/merge and encap/decap and reverse their order

bifurcation:
Currently, the context provided to TreeKEM encap/decap reflects an intermediate ratchet tree, where proposals have been applied, but the sender's direct path has not been updated.  The specification calls for encap/decap to be done [based on the new ratchet tree](https://www.ietf.org/archive/id/draft-ietf-mls-protocol-17.html#section-12.4.1), after both applying proposals and updating the sender's direct path.  This PR splits the `update()` and `encap()` operations, and reverses their order, so that the context provided to encap/decap reflects the new tree.

### DIFF
--- a/src/state.cpp
+++ b/src/state.cpp
@@ -513,11 +513,8 @@ State::commit(const bytes& leaf_secret,
                                              joiner_locations,
                                              leaf_node_opts);
     */
-    auto new_priv = next._tree.update(next._index,
-                                      leaf_secret,
-                                      next._group_id,
-                                      _identity_priv,
-                                      leaf_node_opts);
+    auto new_priv = next._tree.update(
+      next._index, leaf_secret, next._group_id, _identity_priv, leaf_node_opts);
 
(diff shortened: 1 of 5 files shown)


src/state.cpp
```diff
@@ -499,27 +499,33 @@ State::commit(const bytes& leaf_secret,
...
 
+    /*
     auto [new_priv, path] = next._tree.encap(next._index,
                                              next._group_id,
                                              ctx,
                                              leaf_secret,
                                              _identity_priv,
                                              joiner_locations,
                                              leaf_node_opts);
+    */
```
suhasHere, line 515: should we delete the commented code ?

suhasHere (approved):
seems fine to me.

src/state.cpp
bifurcation, line 515: Yep.

```suggestion
```

PR #326, opened 2023-03-10
Support Welcome join scenarios

bifurcation:
In `/src`:
* Remove the unused `encrypt_handshake` flag in `CommitOpts`
* Add a `force_path` flag in `CommitOpts`

In `/cmd/interop`:
* Add support for storing PSKs, both before and after joining


suhasHere (approved):
Looks fine to me

PR #329, opened 2023-03-16
Verify that CI passes on current main

bifurcation:
In #327 there is some indication that the Windows CI build is broken, possibly to the `builtin-baseline` in `vcpkg.json` being too old.  This PR is to verify that this is the problem and hopefully resolve it.


bifurcation:
Error on #327 was transient, unrelated to the failure here.

PR #331, opened 2023-03-17
Support proposals by value

bifurcation:
This PR adds gRPC support for proposals by value.

Corresponds to https://github.com/mlswg/mls-implementations/pull/149

Depends on #328


cmd/interop/src/mls_client_impl.cpp
```diff
@@ -361,12 +361,118 @@ MLSClientImpl::load_state(uint32_t state_id)
...
+    if (basic.identity == id) {
+      break;
+    }
+  }
+
+  if (!(removed_index < tree.size)) {
+    throw std::runtime_error("Unknown member identity");
+  }
+
+  return removed_index;
```
suhasHere, line 431: removed_index may be renamed as member_index ?

suhasHere (approved):

PR #332, opened 2023-03-17
Add support for reinitialization RPCs

bifurcation:
Corresponds to https://github.com/mlswg/mls-implementations/pull/150

Depends on #331


bifurcation:
Oops, this is on the wrong branch!

PR #346, opened 2023-06-12
Fix CI errors due to runner changes

bifurcation:
* Suppress static function errors (`misc-use-anonymous-namespace`)


bifurcation:
Replaced by #350

PR #351, opened 2023-07-20
configure install target for cmake

glhewett:
I have been integrating mlspp into vcpkg, and one of the first steps in doing the integration is the install target.

The install will install includes to a directory `${CMAKE_INSTALL_PREFIX}/mlspp-${MLSPP_VERSION}`

The libraries will be installed into `${CMAKE_INSTALL_PREFIX}/lib`

The cmake configuration files will be installed into `${CMAKE_INSTALL_PREFIX}/share/cmake/mlspp`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -109,7 +109,7 @@ target_link_libraries(${LIB_NAME} bytes tls_syntax hpke)
 target_include_directories(${LIB_NAME}
   PUBLIC
     $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
-    $<INSTALL_INTERFACE:include/${PROJECT_NAME}-${PROJECT_VERSION}>
+    $<INSTALL_INTERFACE:include/${PROJECT_NAME}>
   PRIVATE
     ${OPENSSL_INCLUDE_DIR}
 )
@@ -150,7 +150,7 @@ install(
   DIRECTORY
     include/
   DESTINATION
-    ${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}-${PROJECT_VERSION})
+    ${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME})
 
 install(
   FILES
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,8 @@ option(SANITIZERS "Enable sanitizers" OFF)
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
 include(CheckCXXCompilerFlag)
+include(CMakePackageConfigHelpers)
+include(GNUInstallDirs)
 
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -123,5 +125,38 @@ endif()
...
+write_basic_package_version_file(
+  ${CMAKE_CURRENT_BINARY_DIR}/mlspp-config-version.cmake
+  VERSION ${PROJECT_VERSION}
+  COMPATIBILITY SameMajorVersion)
+
+###
+### Install
+###
+
+install(TARGETS ${LIB_NAME} EXPORT mlspp-targets)
```
bifurcation, line 147: Could we avoid duplication in other CMake files by putting the other targets under `TARGETS` here?

lib/bytes/CMakeLists.txt
```diff
@@ -11,7 +11,21 @@ add_library(${CURRENT_LIB_NAME} ${LIB_HEADERS} ${LIB_SOURCES})
 add_dependencies(${CURRENT_LIB_NAME} tls_syntax)
 target_link_libraries(${CURRENT_LIB_NAME} tls_syntax)
 target_include_directories(${CURRENT_LIB_NAME}
-  PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include
+  PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+    $<INSTALL_INTERFACE:include/${PROJECT_NAME}-${PROJECT_VERSION}>
```
bifurcation, line 16: It seems like this would result in things of the form `#include <mlspp-${VERSION}/header.h>`.  Is that right?  I don't love that pattern; is there a way we could do versioning at a higher level, so that it could be handled at configuration time instead of at the `#include` stage?

CMakeLists.txt
glhewett: I believe that CMAKE configurations make more sense, the closer they are to the code.  That does cause some duplication, but it prevents any one small change from having bigger consequences.

lib/bytes/CMakeLists.txt
glhewett: The version is just window dressing.  It does not make it easier or harder if it there.  It might make it harder if you are looking for a specific version.  There is nothing else in the code that cares about version or supports versions, so I think it is confusing.  If we need to support multiple version, we can address it then.

To answer you question, the following is what the install directory looks like.

```
.
├── mlspp-0.1
│·· ├── bytes
│·· │·· └── bytes.h
│·· ├── hpke
│·· │·· ├── certificate.h
│·· │·· ├── digest.h
│·· │·· ├── hpke.h
│·· │·· ├── random.h
│·· │·· └── signature.h
│·· ├── mls
│·· │·· ├── common.h
│·· │·· ├── core_types.h
│·· │·· ├── credential.h
│·· │·· ├── crypto.h
│·· │·· ├── key_schedule.h
│·· │·· ├── messages.h
│·· │·· ├── session.h
│·· │·· ├── state.h
│·· │·· ├── tree_math.h
│·· │·· └── treekem.h
│·· ├── mls_vectors
│·· │·· └── mls_vectors.h
│·· ├── tls
│·· │·· ├── compat.h
│·· │·· └── tls_syntax.h
│·· ├── variant.hpp
│·· └── version.h
```

glhewett: You can best see this when you install via vcpkg...

```
${PROJECT_DIR}/build/vcpkg_installed/arm64-osx/include/
```

PR #359, opened 2023-08-25
allowing for dynamic top-level namespacing.

glhewett:
Taken from the original mls update. (thanks @suhasHere)

I have added a bit of cmake help that will derive the namespacing version from the PROJECT_VERSION.

```
3.0.0 -> v3
3.1.0 -> v3_1
3.1.1 -> v3_1_1
```


bifurcation:
To add a little more context here, since this is a breaking change: The main thrust of this PR is to add a per-version top-level namespace, which contains all other MLSpp namespaces.  This allows multiple versions of the library to be linked at the same time (e.g., for backward compatibility) without symbol conflicts.

(This is effectively [what Rust/cargo does](https://stackoverflow.com/questions/58739075/how-do-i-import-multiple-versions-of-the-same-crate), but more manually.)

bifurcation (approved):
This looks OK to me, modulo the one CI failure.

My main question is whether the version-based namespacing is a little too fixed and too granular.  If we were willing to assume that downstream consumers were building from source (as opposed to consuming a pre-built binary), then we could just make the namespace / version suffix a build-time option, so that the consumer could set it to whatever they want.  Maybe I'm over-thinking this, though.

PR #360, opened 2023-08-29
UserInfo VC details

bifurcation:
This PR adds more capabilities for UserInfo VC credentials:

* Parsing to extract the salient attributes, including the subject public key
* ... and thus a real implementation of `UserInfoVCCredential::valid_for`
* Validation of the signature on a UserInfo VC given an issuer's public key

Marking as draft right now because:

* [ ] Add tests for VC parsing / validation
* [ ] Rebase on #359

### DIFF
--- a/include/mls/credential.h
+++ b/include/mls/credential.h
@@ -47,23 +47,27 @@ private:
   SignatureScheme _signature_scheme;
 };
 
+tls::ostream&
+operator<<(tls::ostream& str, const X509Credential& obj);
+
+tls::istream&
+operator>>(tls::istream& str, X509Credential& obj);
+
 struct UserInfoVCCredential
 {
   UserInfoVCCredential() = default;
-  explicit UserInfoVCCredential(bytes userinfo_vc_jwt);
+  explicit UserInfoVCCredential(bytes userinfo_vc_jwt_in);
 
   bytes userinfo_vc_jwt;
 
   bool valid_for(const SignaturePublicKey& pub) const;
 
   TLS_SERIALIZABLE(userinfo_vc_jwt)
+
+  private:
+  SignaturePublicKey _public_key;
+  SignatureScheme _signature_scheme;
 };
-
-tls::ostream&
-operator<<(tls::ostream& str, const X509Credential& obj);
-
-tls::istream&
-operator>>(tls::istream& str, X509Credential& obj);
 
 bool
 operator==(const X509Credential& lhs, const X509Credential& rhs);
--- a/lib/hpke/include/hpke/userinfo_vc.h
+++ b/lib/hpke/include/hpke/userinfo_vc.h
@@ -15,22 +15,23 @@ struct UserInfoVC
 {
 private:
   struct ParsedCredential;
-  std::unique_ptr<ParsedCredential> parsed_cred;
+  std::shared_ptr<ParsedCredential> parsed_cred;
 
 public:
   explicit UserInfoVC(std::string jwt);
   UserInfoVC() = delete;
-  UserInfoVC(const UserInfoVC& other);
+  UserInfoVC(const UserInfoVC& other) = default;
   ~UserInfoVC() = default;
 
   std::string issuer() const;
   std::string key_id() const;
-  bool valid_from(const Signature::PublicKey& issuer_key) const;
-
-  // Accessors
-  std::map<std::string, std::string> subject() const;
   std::chrono::system_clock::time_point not_before() const;
   std::chrono::system_clock::time_point not_after() const;
+  std::map<std::string, std::string> subject() const;
+  Signature::ID public_key_algorithm() const;
+  const Signature::PublicKey public_key() const;
+
+  bool valid_from(const Signature::PublicKey& issuer_key) const;
 
   const std::string raw;
 };
--- a/lib/hpke/include/hpke/userinfo_vc.h
+++ b/lib/hpke/include/hpke/userinfo_vc.h
@@ -19,7 +19,7 @@ private:
 
 public:
   explicit UserInfoVC(std::string jwt);
-  UserInfoVC() = delete;
+  UserInfoVC() = default;
   UserInfoVC(const UserInfoVC& other) = default;
   ~UserInfoVC() = default;
 
--- a/lib/hpke/include/hpke/userinfo_vc.h
+++ b/lib/hpke/include/hpke/userinfo_vc.h
@@ -1,0 +1,41 @@
+#pragma once
+#include <memory>
+#include <optional>
+
+#include <bytes/bytes.h>
+#include <chrono>
+#include <hpke/signature.h>
+#include <map>
+
+using namespace bytes_ns;
+
+namespace hpke {
+
+struct UserInfoVC
+{
+private:
+  struct ParsedCredential;
+  std::unique_ptr<ParsedCredential> parsed_cred;
+
+public:
+  explicit UserInfoVC(std::string jwt);
+  UserInfoVC() = delete;
+  UserInfoVC(const UserInfoVC& other);
+  ~UserInfoVC() = default;
+
+  std::string issuer() const;
+  std::string key_id() const;
+  bool valid_from(const Signature::PublicKey& issuer_key) const;
+
+  // Accessors
+  std::map<std::string, std::string> subject() const;
+  std::chrono::system_clock::time_point not_before() const;
+  std::chrono::system_clock::time_point not_after() const;
+
+  const std::string raw;
+};
+
+bool
+operator==(const UserInfoVC& lhs, const UserInfoVC& rhs);
+
+} // namespace hpke
--- a/lib/hpke/src/signature.cpp
+++ b/lib/hpke/src/signature.cpp
@@ -279,4 +279,50 @@ Signature::generate_rsa(size_t bits)
   return RSASignature::generate_key_pair(bits);
 }
 
+static const Signature&
+sig_from_jwk(const std::string& json_str)
+{
+  using KeyTypeAndCurve = std::tuple<std::string, std::string>;
+  static const auto alg_sig_map = std::map<KeyTypeAndCurve, const Signature&>{
+    { { "EC", "P-256" }, Signature::get<Signature::ID::P256_SHA256>() },
+    { { "EC", "P-384" }, Signature::get<Signature::ID::P384_SHA384>() },
+    { { "EC", "P-512" }, Signature::get<Signature::ID::P521_SHA512>() },
+    { { "OKP", "Ed25519" }, Signature::get<Signature::ID::Ed25519>() },
+    { { "OKP", "Ed448" }, Signature::get<Signature::ID::Ed448>() },
+    { { "RSA", "" }, Signature::get<Signature::ID::RSA_SHA256>() },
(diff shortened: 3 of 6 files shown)


lib/hpke/include/hpke/userinfo_vc.h
```diff
@@ -0,0 +1,42 @@
...
+namespace hpke {
+
+struct UserInfoVC
+{
+private:
+  struct ParsedCredential;
+  std::shared_ptr<ParsedCredential> parsed_cred;
+
+public:
+  explicit UserInfoVC(std::string jwt);
```
suhasHere, line 21: why not ``` std::string& jwt ```

lib/hpke/include/hpke/userinfo_vc.h
```diff
@@ -0,0 +1,42 @@
...
+  UserInfoVC() = default;
+  UserInfoVC(const UserInfoVC& other) = default;
+  ~UserInfoVC() = default;
```
suhasHere, line 24: shouldn't we need to define the move and copy assignments too ?

lib/hpke/include/hpke/userinfo_vc.h
```diff
@@ -0,0 +1,42 @@
...
+  std::string key_id() const;
+  std::chrono::system_clock::time_point not_before() const;
+  std::chrono::system_clock::time_point not_after() const;
+  std::map<std::string, std::string> subject() const;
+  Signature::ID public_key_algorithm() const;
+  const Signature::PublicKey public_key() const;
+
+  bool valid_from(const Signature::PublicKey& issuer_key) const;
+
+  const std::string raw;
```
suhasHere, line 36: would s/raw/raw_jwk be more clear?

lib/hpke/src/signature.cpp
```diff
@@ -279,4 +279,50 @@ Signature::generate_rsa(size_t bits)
...
+  const auto jwk = json::parse(json_str);
+  const auto kty = jwk.at("kty");
+
+  auto crv = std::string("");
+  if (jwk.contains("crv")) {
+    crv = jwk.at("crv");
+  }
+
+  const auto key = KeyTypeAndCurve{ kty, crv };
+  return alg_sig_map.at(key);
```
suhasHere, line 306: i wonder what happens when kty is RSA , which one gets returned ?

lib/hpke/src/signature.cpp
```diff
@@ -279,4 +279,50 @@ Signature::generate_rsa(size_t bits)
...
+  const auto& sig = sig_from_jwk(json_str);
+  auto priv = sig.import_jwk_private(json_str);
+  return { sig.id, std::move(priv) };
+}
+
+std::tuple<Signature::ID, std::unique_ptr<Signature::PublicKey>>
+Signature::parse_jwk(const std::string& json_str)
+{
+  // XXX(RLB): Same double-parsing comment as with `parse_jwk_private`
+  const auto& sig = sig_from_jwk(json_str);
```
suhasHere, line 323: agreed , but may be we extract add a utility method that extracts an object with all the details (sig and keys) and return that ?

lib/hpke/src/userinfo_vc.cpp
```diff
@@ -0,0 +1,197 @@
...
+      "UserInfoCredential",
+    };
+    if (vc.at("type") != type) {
+      throw std::runtime_error("malformed VC: incorrect type value");
+    }
+
+    // Parse the subject public key
+    auto credential_subject = vc.at("credentialSubject");
+    const auto id = credential_subject.at("id").get<std::string>();
+    credential_subject.erase("id");
```
suhasHere, line 97: curious, why are we erasing the id here ?

include/mls/credential.h
```diff
@@ -47,23 +47,27 @@ struct X509Credential
...
-};
 
-tls::ostream&
-operator<<(tls::ostream& str, const X509Credential& obj);
-
-tls::istream&
-operator>>(tls::istream& str, X509Credential& obj);
+  private:
+  SignaturePublicKey _public_key;
+  SignatureScheme _signature_scheme;
```
suhasHere, line 69: nit, s/_public_key/public_key and same applies to similar class properties to keep it uniform in the code base ?

lib/hpke/include/hpke/userinfo_vc.h
bifurcation: `jwt` is always copied into `raw`.  Passing by value makes this explicit.

lib/hpke/include/hpke/userinfo_vc.h
bifurcation: It would be `raw_jwt`, but I think that's redundant; a UserInfoVC is always a JWT.

lib/hpke/src/signature.cpp
bifurcation: None, it throws [std::out_of_range](https://en.cppreference.com/w/cpp/error/out_of_range)

lib/hpke/src/signature.cpp
bifurcation: We would just operate on `nlohmann::json`.  The trouble is that to avoid exposing that dependency to the main `mls` module, you would have to have wrapper methods that take `std::string`.  But that's probably what we should do, even though it's a little verbose.

bifurcation: Update: I tried doing this.  It's not really possible to do while both (a) not leaking the nlohmann::json dependency and (b) not inventing a bunch of extra stuff.

lib/hpke/src/userinfo_vc.cpp
```diff
@@ -0,0 +1,197 @@
...
+    static const std::string context = "https://www.w3.org/2018/credentials/v1";
+    if (vc.at("context") != context) {
+      throw std::runtime_error("malformed VC: incorrect context value");
+    }
+
+    static const auto type = std::vector<std::string>{
+      "VerifiableCredential",
+      "UserInfoCredential",
+    };
+    if (vc.at("type") != type) {
```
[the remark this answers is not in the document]
glhewett: Also,it would be nice to get these values from the meta.

glhewett (approved):
Added some thoughts. Most of which probably don't require a response. Other than the fact that I am trying to learn your style.

lib/hpke/src/userinfo_vc.cpp
bifurcation: Because the ID is "consumed" by transforming it into the public key.  I can leave it, though.  Not a big deal.

include/mls/credential.h
bifurcation: This is consistent with `X509Credential`.  We're not super consistent throughout the library, but generally private fields get an underscore prefix.

lib/hpke/src/signature.cpp
```diff
@@ -279,4 +279,50 @@ Signature::generate_rsa(size_t bits)
...
+    { { "OKP", "Ed448" }, Signature::get<Signature::ID::Ed448>() },
+    { { "RSA", "" }, Signature::get<Signature::ID::RSA_SHA256>() },
+    { { "RSA", "" }, Signature::get<Signature::ID::RSA_SHA384>() },
+    { { "RSA", "" }, Signature::get<Signature::ID::RSA_SHA512>() },
+  };
+
+  const auto jwk = json::parse(json_str);
+  const auto kty = jwk.at("kty");
+
+  auto crv = std::string("");
```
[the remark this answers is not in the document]
bifurcation: I like using `auto` and making the initial value explicit.

lib/hpke/src/signature.cpp
```diff
@@ -279,4 +279,50 @@ Signature::generate_rsa(size_t bits)
...
+  static const auto alg_sig_map = std::map<KeyTypeAndCurve, const Signature&>{
+    { { "EC", "P-256" }, Signature::get<Signature::ID::P256_SHA256>() },
+    { { "EC", "P-384" }, Signature::get<Signature::ID::P384_SHA384>() },
+    { { "EC", "P-512" }, Signature::get<Signature::ID::P521_SHA512>() },
+    { { "OKP", "Ed25519" }, Signature::get<Signature::ID::Ed25519>() },
+    { { "OKP", "Ed448" }, Signature::get<Signature::ID::Ed448>() },
+    { { "RSA", "" }, Signature::get<Signature::ID::RSA_SHA256>() },
+    { { "RSA", "" }, Signature::get<Signature::ID::RSA_SHA384>() },
+    { { "RSA", "" }, Signature::get<Signature::ID::RSA_SHA512>() },
+  };
```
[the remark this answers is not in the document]
bifurcation: Good thing to think about.  I thought about pulling this stuff into `jwa.h`, but given that all the usage right now is local to these `.cpp` files, it seems like overkill.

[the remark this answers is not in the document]
bifurcation: We should fix the RSA support, though.  Slightly unclear how best to do that.  Let's do it in a follow-on.

lib/hpke/src/userinfo_vc.cpp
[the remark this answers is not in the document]
bifurcation: Not sure what you mean by "from the meta".

lib/hpke/src/userinfo_vc.cpp
```diff
@@ -0,0 +1,197 @@
...
+    // Parse the components
+    const auto header = json::parse(from_base64url(header_b64));
+    const auto payload = json::parse(from_base64url(payload_b64));
+    const auto to_be_signed = from_ascii(header_b64 + "." + payload_b64);
+    const auto signature = from_base64url(signature_b64);
+
+    // Verify the VC parts
+    const auto vc = payload.at("vc");
+
+    static const std::string context = "https://www.w3.org/2018/credentials/v1";
```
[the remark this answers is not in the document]
bifurcation: No, these constants aren't relevant outside of the process of validating that a VC is correct, i.e., this specific method.

lib/hpke/src/userinfo_vc.cpp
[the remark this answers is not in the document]
glhewett: /.well-known/openid-configuration

lib/hpke/src/userinfo_vc.cpp
```diff
@@ -0,0 +1,197 @@
...
+      "VerifiableCredential",
+      "UserInfoCredential",
+    };
+    if (vc.at("type") != type) {
+      throw std::runtime_error("malformed VC: incorrect type value");
+    }
+
+    // Parse the subject public key
+    auto credential_subject = vc.at("credentialSubject");
+    const auto id = credential_subject.at("id").get<std::string>();
```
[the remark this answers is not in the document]
bifurcation: Given the comment from @suhasHere above about deleting `id`, I streamlined this along another axis.  (removed the `credential_subject` variable altogether.

lib/hpke/src/userinfo_vc.cpp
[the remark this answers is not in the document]
bifurcation: Oh, I see.  In this case, it doesn't matter because these values are fixed by the UserInfo VC spec.

glhewett (approved):
LGTM

You might want to squash the branch. That is just my preference.

suhasHere (approved):

PR #361, opened 2023-08-29
Don't run slow/expensive CI jobs if faster jobs fail

glhewett:
I am looking for a faster CI, and these were some obvious updates to make.  Actually, I could not stop making changes.  Here is a list of the changes in this PR.

1. I have created dependencies between the jobs.  This required consolidating the jobs into one workflow file.  If the clang-format fails, the rest of the job will not run.  Next, I have added the interop job as a gate to do the platform jobs.  This job will not use sanitizers, but it should be a quick test of openssl1, openssl2, and interop.  Once those two jobs succeed, then we will move on to the full platform tests with sanitizer and tidy.
2. I have updated most of the actions to the latest, greatest.
3. I have moved the vcpgk caching to cache the binaries instead of the installed dependencies.  This is a feature of vcpkg and should be very efficient. Since we do not have vcpkg locked as a submodule, then we might have to rebuild some dependencies an update cache from time to time, but vcpkg does not take a long time.  Also, by consolidating the jobs, we have reduced a bit of the vcpkg repetition.
4. We will not run the longer running jobs if it in draft status (not ready_for_review)

Screenshot 2023-08-30 at 11 39 34 AM


bifurcation:
Also it seems like we have Configure/Build/Test everywhere except for the interop harness, which only has Build/Test.  Let's consolidate the Configure/Build steps into just Build.

glhewett:
Great feedback.  I had not taken the higher level view yet.  I have responded to all of you feedback.

PR #362, opened 2023-09-05
Add an equality operator for PublicMessage

bifurcation:
Since PublicMessage uses its own TLS serialization operators instead of `TLS_SERIALIZABLE`, it also needs to define its own equality operator.  This PR adds one, in the obvious way.


bifurcation:
Replaced by #364

PR #363, opened 2023-09-05
Validate the tree on joining

bifurcation:
Currently, on joining, we validate that the tree is tree-hash valid and parent-hash valid, but we do not validate more advanced properties, namely that each leaf is validly signed and each member supports the group's parameters.  This PR adds such checks.

The approach is lightly brute-force.  We leverage the existing method `State::valid()` for checking that a LeafNode is valid in the context of a group.  (Currently used for validating Add proposals and Commits.)  This approach unfortunately scales **quadratically** in the number of members, since each invocation of `State::valid()` causes a linear scan through the tree to check for parameter support and duplicate keys.  This scaling could be reduced to linear at the cost of a bunch of duplicate code for tree validation.

The assumption here is that the quadratic scaling is tolerable because it has reasonable small constants.  The operations involved are non-cryptographic, so they should be reasonably quick.  However, this assumption may turn out to be false at moderately large group sizes; 1k participants will require 1M checks.

While implementing, I noticed that scanning through the leaves of the tree is a common pattern.  So I added a method `TreeKEMPublicKey::all_leaves()` to run a closure on every non-blank leaf, and migrated a bunch of `for` loops over to it.

### DIFF
--- a/src/state.cpp
+++ b/src/state.cpp
@@ -74,19 +74,69 @@ State::import_tree(const bytes& tree_hash,
 bool
 State::validate_tree() const
 {
+  // XXX(RLB): The functionality here is somewhat duplicative of
+  // State::valid(const LeafNode&).  Simply calling that method, however, would
+  // result in this method having quadratic scaling, since each call to valid()
+  // does a linear scan through the tree to check uniqueness of keys and
+  // compatibility of credential support.
+
   // Validate that the tree is parent-hash valid
   if (!_tree.parent_hash_valid()) {
     return false;
   }
 
-  // Validate that each leaf node would be valid to add to the tree now.  This
-  // includes signature verification as well as support for ciphersuites and
-  // extensions.
-  _tree.all_leaves([&](auto i, const auto& leaf_node) {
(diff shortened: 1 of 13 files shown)


src/state.cpp
```diff
@@ -68,11 +68,25 @@ State::import_tree(const bytes& tree_hash,
...
+    return false;
   }
 
-  return tree;
+  // Validate that each leaf node would be valid to add to the tree now.  This
+  // includes signature verification as well as support for ciphersuites and
+  // extensions.
+  _tree.all_leaves([&](auto i, const auto& leaf_node) {
+    return valid(leaf_node, leaf_node.source(), i);
+  });
```
suhasHere, line 87: may be just 
```
return _tree.all_leaves(...) {...}
```

bifurcation:
Update: The validation is now linear-time, at the cost of reproducing some functionality.

bifurcation:
Reviewed IRL with @glhewett yesterday.  Merging despite macOS / OpenSSL3 not finishing because that build seems to hang reliably.

PR #365, opened 2023-09-08
More integration fixes

bifurcation:
This PR makes a few changes

* Aligns the WireFormat enum value names to the RFC
* Updates parent hash computation to properly handle the case when there is only one member in the tree, but the member is not in leaf 0
* Move key uniqueness checking to TreeKEMPublicKey, and remove those checks from proposal validation.


bifurcation:
Reviewed this IRL with @glhewett, agreed it's approved.

PR #367, opened 2023-09-09
adding vcpkg as a submodule to speed up CI

glhewett:
This PR will do two things:

1. Adds vcpkg as a submodule will help us to lock in the version of our dependencies requiring us to never build new dependencies. This is a step in minimizing the changes that require to recapture cache.
2. Reworked the cache key to minimize dependencies rebuilds by using an older cache to prime the new cache.  i.e.  if the vcpkg.json changes, it will use recent relevant cache to start rebuilding a new cache.  (Changes to one of the two vcpkg.json will cause a new cache to be created)

### DIFF
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -153,12 +153,12 @@ jobs:
         git -C ${{ env.VCPKG_REPO }} rev-parse HEAD > vcpkg_commit.txt
 
     - name: Set Deployment Target
-      if: ${{ matrix.os == 'macos-latest' && matrix.deployment_target > 0 }}
+      if: ${{ startsWith(matrix.os, 'macos') && matrix.deployment_target > 0 }}
       run: |
         echo "MACOSX_DEPLOYMENT_TARGET=${{ matrix.deployment_target }}" >> $GITHUB_ENV
 
     - name: Dependencies (macOs)
-      if: ${{ matrix.os == 'macos-latest' }}
+      if: ${{ startsWith(matrix.os, 'macos') }}
       run: |
         brew install llvm pkg-config
         ln -s "/usr/local/opt/llvm/bin/clang-format" "/usr/local/bin/clang-format"
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -29,7 +29,7 @@ jobs:
           - 'cmd'
           - 'lib'
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
 
     - name: Run clang-format style check for C/C++ programs
       uses: jidicula/clang-format-action@v4.11.0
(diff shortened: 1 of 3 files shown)


.github/workflows/main_ci.yml
```diff
@@ -43,22 +43,42 @@ jobs:
...
-        key: VCPKG-BinaryCache-${{ runner.os }}
+        key: VCPKG-BinaryCache-ubuntu-latest-v1-${{ hashFiles('vcpkg-commit.txt', 'vcpkg.json', 'alternatives/openssl_3/vcpkg.json') }}
+        restore-keys: |
+          VCPKG-BinaryCache-ubuntu-latest-v1
+          VCPKG-BinaryCache-ubuntu-latest
 
     - name: Build (OpenSSL 1.1)
       run: |
-        cmake -B "${{ env.CMAKE_BUILD_DIR }}" -DTESTING=ON -DCMAKE_TOOLCHAIN_FILE="$VCPKG_INSTALLATION_ROOT/scripts/buildsystems/vcpkg.cmake"
+        cmake -B "${{ env.CMAKE_BUILD_DIR }}" -DTESTING=ON -DCMAKE_TOOLCHAIN_FILE="vcpkg/scripts/buildsystems/vcpkg.cmake"
```
bifurcation, line 81: You could make these lines a little more concise by making `CMAKE_TOOLCHAIN_FILE` an environment variable (CMake should pick it up from there).  You would need to make it absolute, e.g., by anchoring it on `${{ github.workspace }}`.

.github/workflows/main_ci.yml
```diff
@@ -101,23 +125,36 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [windows-latest, ubuntu-latest, macos-latest]
+        os: [windows-latest, ubuntu-latest, macos-latest, macos-11]
```
bifurcation, line 128: I would note that there's a slightly subtle difference between this setup and the separate build below.  In the current setup, the build platform is `macos-latest`, but with `MACOSX_DEPLOYMENT_TARGET: 10.11`.  Here we build directly on 10.11.  I suspect that the. current setup is more reflective of how people actually build things with compatibility for old platforms (as opposed to keeping around machines actually running the old OS).  But I think this is probably OK in terms of catching incompatibilities / failures.

Makefile
```diff
@@ -18,14 +19,21 @@ all: ${BUILD_DIR}
...
+vcpkg:
+	git submodule update --init --recursive
+
 dev:
 	# Only enable testing, not clang-tidy/sanitizers; the latter make the build
 	# too slow, and we can run them in CI
 	cmake -B${BUILD_DIR} -DTESTING=ON -DCMAKE_BUILD_TYPE=Debug .
 
-dev3:
+vcpkg-dev: vcpkg
```
bifurcation, line 30: I would just update the `dev` target to use vcpkg as well. Might as well be consistent.

Makefile
```diff
@@ -18,14 +19,21 @@ all: ${BUILD_DIR}
 ${BUILD_DIR}: CMakeLists.txt test/CMakeLists.txt
 	cmake -B${BUILD_DIR} .
 
+vcpkg:
+	git submodule update --init --recursive
```
bifurcation, line 23: It would be more explicit to phrase this terms of the toolchain file directly:

```suggestion
${TOOLCHAIN_FILE}:
	git submodule update --init --recursive
```

... and then have the `dev` and `dev3` targets depend on `${TOOLCHAIN_FILE}`.

bifurcation (changes requested):
Overall, this looks pretty good.  Just a few minor things I would like tuned up.

One general comment: It would probably speed things up to set [`VCPKG_BUILD_TYPE`](https://learn.microsoft.com/en-us/vcpkg/users/triplets#vcpkg_build_type) to `release`, which will prevent vcpkg building both debug and release versions of the dependencies.

glhewett:
> Overall, this looks pretty good. Just a few minor things I would like tuned up.
> 
> One general comment: It would probably speed things up to set [`VCPKG_BUILD_TYPE`](https://learn.microsoft.com/en-us/vcpkg/users/triplets#vcpkg_build_type) to `release`, which will prevent vcpkg building both debug and release versions of the dependencies.

I don't think it as simple as just setting an env/cmake variable.  I tried both, and they did not work for me.  I found a stack overflow that suggested creating a new triplet file, which I can look into in the future. Since the cache is working as expected and the downloads are already pretty fast, I don't think this really improve the speed of the builds.

https://stackoverflow.com/questions/52578975/getting-vcpkg-to-build-only-release-version

.github/workflows/main_ci.yml
glhewett: I have hoisted a bunch of variables up in the top env group. check it out

.github/workflows/main_ci.yml
glhewett: I see now. I might have fixed that.

Makefile
glhewett: I don't have a dog in this hunt.  I would consider not requiring the use vcpkg by default.  I have done what you requested.

Makefile
glhewett: done

.github/workflows/main_ci.yml
glhewett: I did fix this, and I was able to open the door to building several macOS targets.

glhewett:
Regarding the `MACOSX_DEPLOYMENT_TARGET`, it looks like it is not working 100%.  I am seeing logs like the following:

```
ld: warning: object file (../../../vcpkg_installed/x64-osx/debug/lib/libcrypto.a(libcrypto-lib-aes-x86_64.o)) was built for newer macOS version (12.7) than being linked (10.11)
ld: warning: object file (../../../vcpkg_installed/x64-osx/debug/lib/libcrypto.a(libcrypto-lib-aes_cfb.o)) was built for newer macOS version (12.7) than being linked (10.11)
```

I search for how I could fix that... 
https://github.com/microsoft/vcpkg/issues/10038

bifurcation (approved):

bifurcation:
Replaced by #395, #391, #389

PR #368, opened 2023-09-09
CI improvements

bifurcation:
* Streamlined format checking, leveraging `include-regex` as well as `check-path`

### DIFF
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -15,6 +15,7 @@ env:
   CMAKE_TEST_OPENSSL3_DIR: ${{ github.workspace }}/build_openssl3/test
   CMAKE_TEST_DIR: ${{ github.workspace }}/build/test
   CMAKE_TOOLCHAIN_FILE: ${{ github.workspace }}/vcpkg/scripts/buildsystems/vcpkg.cmake
+  VCPKG_COMMIT_ID: 3b3bd424827a1f7f4813216f6b32b6c61e386b2e
 
 jobs:
 #  formatting-check:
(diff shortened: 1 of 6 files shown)


.github/workflows/main_ci.yml
```diff
@@ -23,11 +20,8 @@ jobs:
     strategy:
       matrix:
         path:
-          - 'include'
-          - 'src'
-          - 'test'
-          - 'cmd'
-          - 'lib'
+          - '.'
+          - 'cmd/interop'
```
glhewett, line 24: isn't `cmd/interop` a subset of `.`

.github/workflows/main_ci.yml
```diff
@@ -45,30 +39,27 @@ jobs:
...
-        key: VCPKG-BinaryCache-${{ runner.os }}
+        vcpkgGitCommitId:  ${{ env.VCPKG_COMMIT_ID }}
+        vcpkgDirectory: vcpkg
 
     - name: Build (OpenSSL 1.1)
       run: |
-        cmake -B "${{ env.CMAKE_BUILD_DIR }}" -DTESTING=ON -DCMAKE_TOOLCHAIN_FILE="$VCPKG_INSTALLATION_ROOT/scripts/buildsystems/vcpkg.cmake"
-        cmake --build "${{ env.CMAKE_BUILD_DIR }}" --target all --parallel 2
+        cmake -B build -DTESTING=ON
+        cmake --build build --target all
```
glhewett, line 53: I would add the `--parallel 2` back.  Even though we may only have once vcpu, we want to keep the compiler as busy as possible. I have always seen recommendations of cpu.count + 1.

.github/workflows/main_ci.yml
```diff
@@ -83,15 +74,15 @@ jobs:
...
       run: |
-        cmake -B "${{ env.CMAKE_BUILD_OPENSSL3_DIR }}" -DTESTING=ON -DVCPKG_MANIFEST_DIR="alternatives/openssl_3" -DCMAKE_TOOLCHAIN_FILE="$VCPKG_INSTALLATION_ROOT/scripts/buildsystems/vcpkg.cmake"
-        cmake --build "${{ env.CMAKE_BUILD_OPENSSL3_DIR }}"
+        cmake -B build -DTESTING=ON -DVCPKG_MANIFEST_DIR="alternatives/openssl_3"
+        cmake --build build
 
     - name: Unit Test (OpenSSL 3)
       run: |
-        cmake --build "${{ env.CMAKE_BUILD_OPENSSL3_DIR }}" --target test
+        cmake --build build --target test
```
glhewett, line 85: add `--parallel 2`

I won't bore you with adding this repeatedly.

.github/workflows/main_ci.yml
```diff
@@ -45,30 +39,27 @@ jobs:
...
       run: |
-        cmake --build "${{ env.CMAKE_BUILD_DIR }}" --target test
+        cmake --build build --target test
 
     - name: Build (Interop Harness)
       run: |
         cd cmd/interop
-        cmake -B build -DCMAKE_TOOLCHAIN_FILE="$VCPKG_INSTALLATION_ROOT/scripts/buildsystems/vcpkg.cmake"
+        cmake -B build
         cmake --build build
```
glhewett, line 63: add `--paralllel 2`

.github/workflows/main_ci.yml
```diff
@@ -104,86 +95,65 @@ jobs:
...
     steps:
     - uses: actions/checkout@v3
 
-    - name: Dependencies (macOs)
-      if: ${{ matrix.os == 'macos-latest' }}
-      run: |
-        brew install llvm pkg-config
-        ln -s "/usr/local/opt/llvm/bin/clang-format" "/usr/local/bin/clang-format"
-        ln -s "/usr/local/opt/llvm/bin/clang-tidy" "/usr/local/bin/clang-tidy"
+    - uses: lukka/get-cmake@latest
```
glhewett, line 110: for a build environment, I would try and lock in a version.

```diff
- uses: lukka/get-cmake@latest
+ uses: lukka/get-cmake@3.26.4
```

alternatives/openssl_3/vcpkg.json
```diff
@@ -10,7 +10,7 @@
     "doctest",
     "nlohmann-json"
   ],
-  "builtin-baseline": "5908d702d61cea1429b223a0b7a10ab86bad4c78",
+  "builtin-baseline": "962e5e39f8a25f42522f51fffc574e05a3efd26b",
```
glhewett, line 13: If this passes CI, that is great!

glhewett (approved):
I have one recommendation, which is not critical. Otherwise, LGTM

glhewett (changes requested):
I am seeing some warnings from the running jobs that might need to be addressed.  One of them is regarding caching.

https://github.com/cisco/mlspp/actions/runs/6132855786

.github/workflows/main_ci.yml
bifurcation: No, because the `include-regex` limits makes each of these only apply to `src|include|lib|test`.

bifurcation:
Closing this in favor of #367.  Some of the cleanup here may still be relevant, but we should get that PR tuned up and merged and then revisit.

PR #369, opened 2023-09-10
provide the user access to the credentialSubject via UserInfoClaims structure

glhewett:
This PR will change the format of the credentialSubject from a map of string to a struct that contains the claims and defined in the specification.

https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims

### DIFF
--- a/lib/hpke/include/hpke/userinfo_vc.h
+++ b/lib/hpke/include/hpke/userinfo_vc.h
@@ -11,6 +11,71 @@ using namespace MLS_NAMESPACE::bytes_ns;
 
 namespace MLS_NAMESPACE::hpke {
 
+struct UserInfoClaims
+{
+  static inline const std::string name_attr = "name";
+  static inline const std::string sub_attr = "sub";
+  static inline const std::string given_name_attr = "given_name";
+  static inline const std::string family_name_attr = "family_name";
+  static inline const std::string middle_name_attr = "middle_name";
+  static inline const std::string nickname_attr = "nickname";
+  static inline const std::string preferred_username_attr =
+    "preferred_username";
+  static inline const std::string profile_attr = "profile";
+  static inline const std::string picture_attr = "picture";
+  static inline const std::string website_attr = "website";
+  static inline const std::string email_attr = "email";
+  static inline const std::string email_verified_attr = "email_verified";
+  static inline const std::string gender_attr = "gender";
+  static inline const std::string birthdate_attr = "birthdate";
+  static inline const std::string zoneinfo_attr = "zoneinfo";
+  static inline const std::string locale_attr = "locale";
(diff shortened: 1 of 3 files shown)


lib/hpke/include/hpke/userinfo_vc.h
```diff
@@ -29,7 +94,7 @@ struct UserInfoVC
   std::string key_id() const;
   std::chrono::system_clock::time_point not_before() const;
   std::chrono::system_clock::time_point not_after() const;
-  std::map<std::string, std::string> subject() const;
+  std::shared_ptr<UserInfoClaims> subject() const;
```
bifurcation, line 97:
```suggestion
  const UserInfoClaims& subject() const;
```

lib/hpke/src/userinfo_vc.cpp
```diff
@@ -130,7 +130,7 @@ struct UserInfoVC::ParsedCredential
   std::chrono::system_clock::time_point not_after;  // `exp`
 
   // Credential subject fields
-  std::map<std::string, std::string> credential_subject;
+  std::shared_ptr<UserInfoClaims> credential_subject;
```
bifurcation, line 133:
```suggestion
  UserInfoClaims credential_subject;
```

lib/hpke/src/userinfo_vc.cpp
```diff
@@ -142,7 +142,7 @@ struct UserInfoVC::ParsedCredential
                    std::string issuer_in,
                    std::chrono::system_clock::time_point not_before_in,
                    std::chrono::system_clock::time_point not_after_in,
-                   std::map<std::string, std::string> credential_subject_in,
+                   std::shared_ptr<UserInfoClaims> credential_subject_in,
```
bifurcation, line 145:
```suggestion
                   UserInfoClaims credential_subject_in,
```

lib/hpke/src/userinfo_vc.cpp
```diff
@@ -222,7 +222,7 @@ struct UserInfoVC::ParsedCredential
       epoch_time(payload.at("nbf").get<int64_t>()),
       epoch_time(payload.at("exp").get<int64_t>()),
 
-      vc.at("credentialSubject"),
+      UserInfoClaims::from_json(vc.at("credentialSubject").dump()),
```
bifurcation, line 225: If you define `to_json / from_json` for `UserInfoClaims, you should be able to just pass `vc.at("credentialSubject")`.

https://github.com/nlohmann/json#basic-usage

glhewett: that would require the end-user to using nlohmann/json.

bifurcation (approved):

PR #370, opened 2023-09-11
Provide more flexible JWK parsing

bifurcation:
Currently, JWK import requires the caller to specify the cipher suite for the key.  This is not always known.  This PR introduces a `PublicJWK` class that provides more fulsome JWK parsing, including capturing the `alg` and `kid` parameters.

### DIFF
--- a/include/mls/credential.h
+++ b/include/mls/credential.h
@@ -6,6 +6,10 @@
 
 namespace MLS_NAMESPACE {
 
+namespace hpke {
+  struct UserInfoVC;
+}
+
 // struct {
 //     opaque identity<0..2^16-1>;
 //     SignaturePublicKey public_key;
@@ -57,17 +61,20 @@ operator>>(tls::istream& str, X509Credential& obj);
 struct UserInfoVCCredential
 {
   UserInfoVCCredential() = default;
-  explicit UserInfoVCCredential(bytes userinfo_vc_jwt_in);
+  explicit UserInfoVCCredential(std::string userinfo_vc_jwt_in);
 
-  bytes userinfo_vc_jwt;
+  std::string userinfo_vc_jwt;
 
   bool valid_for(const SignaturePublicKey& pub) const;
+  bool valid_from(const PublicJWK& pub) const;
 
-  TLS_SERIALIZABLE(userinfo_vc_jwt)
+  friend tls::ostream operator<<(tls::ostream& str, const UserInfoVCCredential& obj);
+  friend tls::istream operator>>(tls::istream& str, UserInfoVCCredential& obj);
+  friend bool operator==(const UserInfoVCCredential& str, const UserInfoVCCredential& obj);
+  friend bool operator!=(const UserInfoVCCredential& str, const UserInfoVCCredential& obj);
 
 private:
-  SignaturePublicKey _public_key;
-  SignatureScheme _signature_scheme;
+  std::shared_ptr<hpke::UserInfoVC> _vc;
 };
 
 bool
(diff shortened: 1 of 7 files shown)


include/mls/crypto.h
```diff
@@ -208,6 +208,8 @@ extern const std::string group_info;
 extern const std::string multi_credential;
 } // namespace sign_label
 
+struct PublicJWK;
```
glhewett, line 211: ❓ Is this necessary

include/mls/credential.h
```diff
@@ -57,17 +61,23 @@ operator>>(tls::istream& str, X509Credential& obj);
...
+  bool valid_from(const PublicJWK& pub) const;
 
-  TLS_SERIALIZABLE(userinfo_vc_jwt)
+  friend tls::ostream operator<<(tls::ostream& str,
+                                 const UserInfoVCCredential& obj);
+  friend tls::istream operator>>(tls::istream& str, UserInfoVCCredential& obj);
+  friend bool operator==(const UserInfoVCCredential& lhs,
+                         const UserInfoVCCredential& rhs);
+  friend bool operator!=(const UserInfoVCCredential& lhs,
+                         const UserInfoVCCredential& rhs);
```
glhewett, line 77: 👍 The macro is nice, but I understand the desire to do std:string for the the json based parameters.

lib/hpke/src/userinfo_vc.cpp
```diff
@@ -223,6 +223,11 @@ struct UserInfoVC::ParsedCredential
       signature = jws_to_der_sig(signature);
     }
 
+    auto kid = std::optional<std::string>{};
+    if (header.contains("kid")) {
+      kid = header.at("kid").get<std::string>();
+    }
```
glhewett, line 229: ❓ I will keep reading... but, I don't understand how kid can be optional.  I was under the impression that several public keys from the issuer might be used.  I am not sure that the test client has the ability to handle the case without kid.

src/credential.cpp
```diff
@@ -115,22 +115,57 @@ operator==(const X509Credential& lhs, const X509Credential& rhs)
...
+bool
+UserInfoVCCredential::valid_from(const PublicJWK& pub) const
+{
+  const auto& sig = _vc->signature_algorithm();
+  if (pub.signature_scheme != tls_signature_scheme(sig.id)) {
+    return false;
+  }
+
+  const auto sig_pub = sig.deserialize(pub.public_key.data);
+  return _vc->valid_from(*sig_pub);
```
glhewett, line 141: ❓ 

Should this be return `_vc->valid_for()`?  Looks like it is s SignaturePublicKey instead of a PublicJWK.  Also, that would result in infinite recursion.

glhewett (commented):
Looks good to me.  I have one comment that requires review.

I also see in this PR and in other places the comparison of json string to provide equality between to objects (Credential or Keys).  The way that we are using the objects, I don't think there is an issue, but there is the case that attributes from a json string can change order and break the equality erroneously.  It might not be a real case, but I want to note it.

include/mls/credential.h
bifurcation: Yeah, it avoids a bunch of copying back and forth between `std::string` and `bytes`.

lib/hpke/src/userinfo_vc.cpp
bifurcation: The `kid` field is OPTIONAL in the [JWK RFC](https://datatracker.ietf.org/doc/html/rfc7517#section-4.5).  There are two plausible modes for a JWT signer to operate in:

1. No `kid`, only one key in the JWK Set 
2. Multiple keys in the JWK Set, use `kid` to distinguish

The specs are silent on which a signer should do, so we should support both options.

src/credential.cpp
bifurcation: No, `valid_from()` is correct.

This is confusing because we have multiple types of signature public key around.  The type of `*sig_pub` is `Signature::PublicKey`, so we're not recursing.  We're calling `UserInfoVC::valid_from(const Signature::PublicKey&)` from `UserInfoVCCredential::valid_from(const PublicJWK&)`, basically moving down a layer of abstraction.

That said, I don't love the number of layers of abstraction here.  We should probably clean this up in a follow-on.

PR #371, opened 2023-09-13
keep nlohmann::json private

glhewett:


bifurcation (approved):

bifurcation:
@glhewett Could you make a trivial push here so that the full CI runs, now that it's not marked as draft?

glhewett:
@bifurcation - It is not working, yet.  I would rather close the PR.  You have my permission to close it once you read this.  I will make sure we have an issue to track it.

lib/hpke/test/userinfo_vc.cpp
```diff
@@ -214,14 +215,16 @@ TEST_CASE("UserInfoClaims Field Parsing")
 TEST_CASE("UserInfoClaims Edge Cases")
 {
   CHECK_THROWS_WITH(
-    UserInfoClaims::from_json({ { "updated_at", "42" } }),
+    UserInfoClaims::from_json(
+      nlohmann::json({ { "updated_at", "42" } }).dump()),
```
bifurcation, line 219: Instead of constructing and dumping, you could just have the JSON strings here, using C++ raw strings to avoid quote ugliness.

```suggestion
    UserInfoClaims::from_json(R"({"updated_at", "42"})"),
```

lib/hpke/src/userinfo_vc.cpp
```diff
@@ -298,13 +300,15 @@ UserInfoClaimsAddress::from_json(const nlohmann::json& address_json)
...
 {
+  const auto& cred_subject_json = nlohmann::json::parse(cred_subject);
+
   std::optional<UserInfoClaimsAddress> address_opt = {};
 
   if (cred_subject_json.contains(address_attr)) {
-    address_opt =
-      UserInfoClaimsAddress::from_json(cred_subject_json.at(address_attr));
+    address_opt = UserInfoClaimsAddress::from_json(
+      cred_subject_json.at(address_attr).dump());
```
bifurcation, line 311: I don't love this back and forth between strings and JSON.  Maybe there's some way to have static functions that take `json`?  So that they're hidden from callers but we can still use them.

bifurcation (approved):
Couple of minor suggestions, otherwise LGTM.

lib/hpke/src/userinfo_vc.cpp
glhewett: This is the difficult thing about this library. If we have any use of our json library in the public api's then the consumer will have to build and link to our json library. If they use another library then they would have to convert. I think you understand that. I will think about the static function that might do the conversion. Out only options in the signature of a function is string or bytes. what do you have in mind?

lib/hpke/test/userinfo_vc.cpp
```diff
@@ -175,7 +175,8 @@ TEST_CASE("UserInfoClaims Field Parsing")
     { "updated_at", 42 }
   };
 
-  const auto userinfo_claims = UserInfoClaims::from_json(credentialSubject);
+  const auto userinfo_claims =
+    UserInfoClaims::from_json(credentialSubject.dump());
```
bifurcation, line 179: Same raw string trick here.  (Raw strings can span multiple lines.)

lib/hpke/src/userinfo_vc.cpp
bifurcation: What I had in mind was something like:

```
// Standard from_json method so that `json::get` works
static void from_json(const json& j, UserInfoClaimsAddress& addr) {
  // current body of UserInfoClaimsAddress::from_json
}

static void from_json(const json& j, UserInfoClaims& addr) { 
  // current body of UserInfoClaims::from_json
  // including using get_optional<UserInfoClaimsAddress>(...)
}

// String methods just call through to JSON methods
UserInfoClaims
UserInfoClaims::from_json(const std::string& cred_subject) {
  return json::parse(cred_subject).get<UserInfoClaims>();
}

UserInfoClaimsAddress
UserInfoClaimsAddress::from_json(const std::string& address) {
  return json::parse(address).get<UserInfoClaimsAddress>();
}
```

lib/hpke/test/userinfo_vc.cpp
glhewett: I am using the original json as the known answer for the  tests.  The tests in that section seem to only test that the strings get shuffled around correctly.

glhewett:
created #393 to optimize implementation

PR #374, opened 2023-09-26
add missing include for inserter

birarda:
Explicitly include the std library `iterator` header in `common.h`.

In some standard library implementations the surrounding headers in `common.h` happen to include it (specifically `array` in one case I checked).  When this is not the case the build fails to find `inserter` in the `std` namespace.


bifurcation (approved):

bifurcation:
macOS + OpenSSL3 CI timeout is a known issue.

PR #376, opened 2023-10-09
Optional replacement of OpenSSL link with BoringSSL

birarda:
If `REQUIRE_BORINGSSL` is passed, requires that the found "OpenSSL" is BoringSSL. The found BoringSSL is subject to the same OpenSSL version requirements (based on the OpenSSL it is said to be compatible with).

It is expected that mlspp consumers will need to supply `OPENSSL_ROOT_DIR` and possibly `OPENSSL_INCLUDE_DIR`, `OPENSSL_CRYPTO_LIBRARY`, and `OPENSSL_SSL_LIBRARY` to point CMake's FindOpenSSL module to their BoringSSL installation. Using vcpkg makes this all easier, and that is what is done in CI.


birarda:
This doesn't yet have the workflow we discussed, working on that now.

CMakeLists.txt
```diff
@@ -10,6 +10,8 @@ option(CLANG_TIDY "Perform linting with clang-tidy" OFF)
 option(SANITIZERS "Enable sanitizers" OFF)
 option(MLS_NAMESPACE_SUFFIX "Namespace Suffix for CXX and CMake Export")
 
+set(BORINGSSL_ROOT_DIR "" CACHE STRING "Root of BoringSSL directory (non-empty string forces use of BoringSSL)")
```
birarda, line 13: This isn't required, but I think it'll highlight the option in the GUI with a helpful description.

CMakeLists.txt
```diff
@@ -91,18 +95,44 @@ endif()
...
+   endif ()
+
+   set(OPENSSL_ROOT_DIR ${BORINGSSL_ROOT_DIR})
+   set(OPENSSL_INCLUDE_DIR ${BORINGSSL_INCLUDE_DIR})
+   set(OPENSSL_CRYPTO_LIBRARY ${BORINGSSL_CRYPTO_LIBRARY})
+   set(OPENSSL_SSL_LIBRARY ${BORINGSSL_SSL_LIBRARY})
+   
+   find_package(OpenSSL REQUIRED)
+
+   set(OPENSSL_VERSION ${BORINGSSL_COMPAT_VERSION})
```
birarda, line 120: We need to override this after calling `find_package(OpenSSL REQUIRED)` since it gets blanked out otherwise.

birarda:
@bifurcation what spacing would you like in the CMake files?

PR #377, opened 2023-10-11
allow CMake/compile option to DISABLE_GREASE

birarda:
Adds a non-default behaviour to disable greasing of extensions/capabilities. This behaviour is recommended as something you should do by the MLS protocol, but it would be nice for a library consumer to be able to disable it as it can reduce the size of messages that need to be passed around.

I went back and forth on making the definition `ADD_GREASE` or `DISABLE_GREASE`, open to changing it to the "positive" direction so we can avoid the `#ifndef DISABLE_` pattern if it is bothersome.

### DIFF
--- a/src/grease.cpp
+++ b/src/grease.cpp
@@ -6,6 +6,21 @@
 
 namespace MLS_NAMESPACE {
 
+#ifdef DISABLE_GREASE
+
+Capabilities
+grease(Capabilities&& capabilities, [[maybe_unused]] const ExtensionList& extensions)
+{
+  return capabilities;
+}
+
+ExtensionList
+grease(ExtensionList&& extensions) {
+  return extensions;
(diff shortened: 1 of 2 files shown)


src/grease.cpp
```diff
@@ -6,6 +6,21 @@
...
+Capabilities
+grease(Capabilities&& capabilities, [[maybe_unused]] const ExtensionList& extensions)
+{
+  return capabilities;
+}
+
+ExtensionList
+grease(ExtensionList&& extensions) {
+  return extensions;
+}
```
birarda, line 20: I had started by `ifdef-ing` each call to `grease(...)` but that creates the branching in many places. This limits the scope of the changes required at the cost of keeping calls to `grease` around that do nothing.

Issue #380, opened 2023-10-24
MacOs runners using newer version of clang-tidy which breaks CI


glhewett:
It seems like the newer versions of LLVM (clang-tidy) have tests that do not pass on the current mlspp source.

```
/Users/runner/work/mlspp/mlspp/lib/tls_syntax/src/tls_syntax.cpp:8:31: error: no header providing "std::vector" is directly included [misc-include-cleaner,-warnings-as-errors]
    3 | 
    4 | // NOLINTNEXTLINE(llvmlibc-implementation-in-namespace)
    5 | namespace MLS_NAMESPACE::tls {
    6 | 
    7 | void
    8 | ostream::write_raw(const std::vector<uint8_t>& bytes)
      |                               ^
/Users/runner/work/mlspp/mlspp/lib/tls_syntax/src/tls_syntax.cpp:8:38: error: no header providing "uint8_t" is directly included [misc-include-cleaner,-warnings-as-errors]
    1 | #include <namespace.h>
    2 | #include <tls/tls_syntax.h>
    3 | 
    4 | // NOLINTNEXTLINE(llvmlibc-implementation-in-namespace)
    5 | namespace MLS_NAMESPACE::tls {
    6 | 
    7 | void
    8 | ostream::write_raw(const std::vector<uint8_t>& bytes)
      |                                      ^
/Users/runner/work/mlspp/mlspp/lib/tls_syntax/src/tls_syntax.cpp:16:21: error: no header providing "uint64_t" is directly included [misc-include-cleaner,-warnings-as-errors]
    1 | #include <namespace.h>
    2 | #include <tls/tls_syntax.h>
    3 | 
    4 | // NOLINTNEXTLINE(llvmlibc-implementation-in-namespace)
    5 | namespace MLS_NAMESPACE::tls {
    6 | 
    7 | void
    8 | ostream::write_raw(const std::vector<uint8_t>& bytes)
    9 | {
   10 |   // Not sure what the default argument is here
   11 |   _buffer.insert(_buffer.end(), bytes.begin(), bytes.end());
   12 | }
   13 | 
   14 | // Primitive type writers
   15 | ostream&
   16 | ostream::write_uint(uint64_t value, int length)
      |                     ^
```

glhewett:
fixed by #385

PR #385, opened 2023-10-24
Fix lints flagged by llvm 17

bifurcation:
Work on #367 indicated that the latest `llvm` added some lints that were hitting on our code.  This PR updates to address those lints:

* Forwarding parameters are either converted to `const&` or passed to `std::forward`
* GREASE is refactored to operate on non-const references instead of moving in and out
* Some lints are ignored (e.g., misc-include-cleaner)


glhewett:
fixes #380

.clang-tidy
```diff
@@ -20,12 +20,15 @@ Checks:            '*,
...
+                    -llvmlibc-inline-function-decl,
                     -llvmlibc-restrict-system-libc-headers,
+                    -misc-include-cleaner,
+                    -misc-no-recursion,
                     -misc-non-private-member-variables-in-classes,
                     -misc-use-anonymous-namespace,
-                    -misc-no-recursion,
                     -modernize-use-nodiscard,
                     -modernize-use-trailing-return-type,
+                    -performance-avoid-endl,
```
glhewett, line 31: Some of these clang-tidy exclusions look handy, but I am sure they are a pain.  I am not going to look at every one.

glhewett (approved):
LGTM

.clang-tidy
bifurcation: We can come back and review them later.

bifurcation:
Merging despite macOS hang, known issue.

PR #389, opened 2023-10-24
adds vcpkg as a submodule

### LINKED ISSUE #381: Add vcpkg as a submodule to lock commit
When vcpkg is added to the project as a submodule, it will lock the commit of vcpkg that the project is using. This will prevent vcpkg from creating surprises that break the build.  It will also provide users with a good experience since they will be using the same version out of the box that was verified in CI.

Note: The use of vcpkg will remain optional.

glhewett:
- updates Makefile with convenience targets including initializing the submodule if it is not there.
- updates build CI with hoisted, constant VCPKG_TOOLCHAIN_FILE environmental variable

Fixes #381


bifurcation (approved):

bifurcation:
Thanks for pulling out this subset of the changes from #367.  That one was getting pretty complex.

bifurcation:
Merging despite macOS hang, known issue.

PR #391, opened 2023-10-25
maximize parallelization

### LINKED ISSUE #384: Maximize parallelization
Each of the runners run with multiple cores.  We should maximize the use of all of the cores.  In the document below, the number of cores are documented.  Each build step should have a `--parallel` argument to specify the number of cores to build with.

From the information provide, we should target having (number of cores + 1) threads.  Since linux and windows only have 2 cores, we should use 3 parallel executions.  MacOs will utilize all 3 of its cores with that value as well.

ref: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

glhewett:
fixes #384

### DIFF
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -9,6 +9,7 @@ on:
       - main
 
 env:
+  CMAKE_GENERATOR: Ninja
   CTEST_OUTPUT_ON_FAILURE: 1
   CMAKE_BUILD_DIR: ${{ github.workspace }}/build
   CMAKE_BUILD_OPENSSL3_DIR: ${{ github.workspace }}/build_openssl3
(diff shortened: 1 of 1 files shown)


bifurcation (commented):
Why is three the right number?  Maybe instead of manually enabling parallelism, we should just set `CMAKE_GENERATOR=Ninja`, which automatically builds in parallel with the right number of jobs.

bifurcation:
I notice some unrelated CI failures in the Ubuntu build.  Not sure where those came from, but they shouldn't be covered in this PR.  I'll do a separate PR to fix them.

glhewett:
ubuntu and windows have 2 cores.... macOS has 3 cores.

Setting the parallel jobs to 3 will be best for all runners.

PR #399, opened 2023-10-27
Simplify clang-format CI job

bifurcation:
For quite some time, the clang-format CI job has run as a matrix across several individual directories.  This clutters up the GitHub interface, since GitHub considers each entry in the matrix a different job.  This PR removes the matrix run and instead uses the `include-regex` feature of the clang-format job to select which files to look at.  In case the CI output doesn't show it, I verified locally that the regex here hits the right files.


glhewett (approved):
LGTM

bifurcation:
Merging in advance of the other CI runners completing because nothing is modified here that would cause them to fail.

PR #400, opened 2023-10-27
CI Cleanup

bifurcation:
Bunch of small fixes to make CI more rational / elegant:

* Make CMAKE_TOOLCHAIN_FILE an environment variable
* Make OpenSSL 1.1 parallel to other crypto libraries
* Simplify out the `vcpkg-dir` variables

Depends on #398


.github/workflows/main_ci.yml
```diff
@@ -9,18 +9,23 @@ on:
...
+
   VCPKG_BINARY_SOURCES: files,${{ github.workspace }}/build/cache,readwrite
-  VCPKG_TOOLCHAIN_FILE: ${{ github.workspace }}/vcpkg/scripts/buildsystems/vcpkg.cmake
   VCPKG_REPO: ${{ github.workspace }}/vcpkg
   CACHE_VERSION: v01
   CACHE_NAME: vcpkg
 
+  OPENSSL_11_VCPKG: alternatives/openssl_1.1
+  OPENSSL_3_VCPKG: alternatives/openssl_3
+  BORINGSSL_VCPKG: alternatives/boringssl
```
glhewett, line 27: I find myself getting confused about which vcpkg attribute we might be talking about.  I would recommend renaming this to `*_VCPKG_MANIFEST` to match the cmake option.

.github/workflows/main_ci.yml
```diff
@@ -196,7 +215,7 @@ jobs:
     runs-on: macos-latest
 
     env:
-        CMAKE_BUILD_DIR: ${{ github.workspace }}/build
+        BUILD_DIR: ${{ github.workspace }}/build
         VCPKG_BINARY_SOURCES: files,${{ github.workspace }}/build/cache,readwrite
```
glhewett, line 219: I know you did not change anything, but this is redundant. It should be at the top.

glhewett (approved):

glhewett (commented):
I think it would be wise to have a default vcpkg.json in the root of the project. Not because it is useful but because it what people expect.

Alternatively, when we add the build directions, we can request the user to copy or symlink the file.  We might want to add it to the .gitignore if that is the case.... just a brainstorm

.github/workflows/main_ci.yml
```diff
@@ -34,190 +34,139 @@ jobs:
...
-    needs: quick-linux-interop-check
-    name: Build for older MacOS
-    runs-on: macos-latest
+    needs: build-and-unit-test
+    name: Build with clang-tidy
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        crypto: [openssl_1.1, openssl_3, boringssl]
```
glhewett, line 166: It might clean things up to make this the vcpkg manifest directory instead of crypto.  It would make it more generic.  Maybe it is a premature optimization.  Just wanted to mention.

bifurcation:
Closed in favor of #406 406

PR #402, opened 2023-10-29
Removes the `from_json()` call for `UserInfoVCAddress`

### LINKED ISSUE #393: from_json methods in UserInfoVC can be optimized to single parse
When importing UserInfo it looks like we are parsing the address twice.  We should remove that parse.

glhewett:
In it's first implementation, it was dumping and parsing the UserInfoVCAddress json structure twice.

This PR will remove the ability to import an UserInfoVCAddress directly. Importing of UserInfoVC will also import the UserInfoVCAddress.

As far as I know, there is not a use case to import the Address directory.  Nor were then any tests.

fixes #393


glhewett:
did not mean to update vcpkg

PR #404, opened 2023-10-30
Make zeroization test tighter and more reliable

bifurcation:
Currently, the zeroization test fails if it finds any bytes that matches the canary.  This is both incorrect, since we should be checking that the vector is all-zero, and failure prone, since it will match against random data with some probability.  This PR changes the test to verify that basically all of the vector is zero, with some allowance for the allocator to re-use a few bytes.

Fully solving this problem would require defining a new allocator (so that we could control what happens on deallocation, and avoid UAF) and updating `bytes` so that it could be provided a non-default allocator.  However, this is a pretty non-trivial change, and not really worth it to fix this one test.

### DIFF
--- a/lib/bytes/test/bytes.cpp
+++ b/lib/bytes/test/bytes.cpp
@@ -12,7 +12,7 @@ using namespace std::literals::string_literals;
 #ifndef SANITIZERS
 TEST_CASE("Zeroization")
 {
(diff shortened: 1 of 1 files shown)


bifurcation:
Apparently the Linux allocator uses more of the freed buffer than macOS!

PR #406, opened 2023-11-02
Refactor CI to return better results faster

bifurcation:
This PR replaces #398, which started out small and turned into kind of a disaster.  This PR restructures the MLSpp CI.  Before, we did a single Linux build first and then ran clang-tidy build on all three platforms:

image

With this PR, we immediately launch build + unit test jobs on all three platforms, with all three choices of crypto library.  Clang-tidy is done only on Ubuntu (the most available platform, and unlike macOS, doesn't hang), and interop testing is only done once (also using Ubuntu).

image

Along the way, I made some smaller changes that make the YAML easier to read:

* Make CMAKE_TOOLCHAIN_FILE an environment variable
* Make OpenSSL 1.1 parallel to other crypto libraries
* Simplify out the vcpkg-dir variables
* Factor out build preparation steps into a composite action

The only change to code is a minor change to zeroization and its test in `lib/bytes/`.  The Zeroization test in that file deliberately reads freed memory in an attempt to verify that memory is zeroized before being freed.  This test was already disabled when ASAN was enabled (since its whole purpose is to detect UAF!), but we also disabled on Windows because it seemed to be causing problems.  While I was researching Windows memory issues, I discovered the `SecureZeroMemory()` function, which seemed safer than the `std::fill()` call that was there before, so I added code to use that.

Nonetheless, Windows CI builds still seem to hang during the unit tests when they are built without `SANITIZERS=ON`, so that flag is always on in the CI builds, which slows them down a little.

My only real remaining question is whether we should continue to run the different crypto libraries matrix-style, or whether we should just run the three of them manually in one job.  I find the 9 jobs a little aesthetically unpleasing, and caching might work a little better.


birarda:
> My only real remaining question is whether we should continue to run the different crypto libraries matrix-style, or whether we should just run the three of them manually in one job. I find the 9 jobs a little aesthetically unpleasing, and caching might work a little better.

I prefer the 9 jobs personally. Definitely a bit unwieldy in the UI to deal with more tasks, but I think it's nice to isolate since it makes clear at the top level what is failing or what is yet to be run. There might also be some parallelization benefits to having them separate? Not familiar enough with actions to know how that works.

.github/actions/build/action.yml
```diff
@@ -0,0 +1,36 @@
...
+      run: |
+         git -C vcpkg rev-parse HEAD > vcpkg_commit.txt
+
+    - name: Restore cache
+      uses: actions/cache@v3
+      with:
+        path: ${{ inputs.cache-dir }}
+        key: v01-vcpkg-${{ runner.os }}-${{ hashFiles('vcpkg_commit.txt', 'alternatives/*/vcpkg.json') }}
+        restore-keys: |
+          v01-vcpkg-${{ runner.os }}
```
glhewett, line 22: This will probably need some more work.  

1. I personally do not favor the runner.os because it is not specific enough to provide proper vcpkg caching.  We can solve this when we observe the failure, but I believe there will be cases where the windows-2022 and windows-2019 might not be able to share the same cache. (VCPKG has its own caching strategy).  Again, let's not sweat this until we see it.
2. We probably need to solve this before merge... Since the first writer will win in creating the cache, we will need to add the crypto library as a variable in the the cache key.  Are you seeing this fail?  If not, I am surprised.

.github/workflows/main_ci.yml
```diff
@@ -34,226 +28,116 @@ jobs:
...
     strategy:
-      fail-fast: false
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
+        crypto: [openssl_1.1, openssl_3, boringssl]
         include:
             - os: windows-latest
-              ossl3-vcpkg-dir: "alternatives\\openssl_3"
-              boringssl-vcpkg-dir: "alternatives\\boringssl_1.1"
               ctest-target: RUN_TESTS
```
glhewett, line 41: Even thought it works, I am not a bit fan us using the matrix to hold variables.  I think it is confusing.  I think there should be a top level variable... `CMAKE_TEST_TARGET_DEFAULT` and `CMAKE_TEST_TARGET_WINDOWS` - Yes, it would be an conditional step, but you would be able to remove the confusing include section of the matrix.

.github/workflows/main_ci.yml
```diff
@@ -34,226 +28,116 @@ jobs:
...
-              boringssl-vcpkg-dir: "alternatives/boringssl_1.1"
               ctest-target: test
             - os: macos-latest
-              ossl3-vcpkg-dir: "alternatives/openssl_3"
-              boringssl-vcpkg-dir: "alternatives/boringssl_1.1"
               ctest-target: test
 
+    env:
+      BUILD_DIR: "${RUNNER_TEMP}/build_${{ matrix.crypto }}"
+      CRYPTO_DIR: "./alternatives/${{ matrix.crypto }}"
```
glhewett, line 49: 💯

.github/workflows/main_ci.yml
```diff
@@ -34,226 +28,116 @@ jobs:
...
-    if: github.event.pull_request.draft == false
-    needs: quick-linux-interop-check
-    name: Build and test platforms using sanitizers and clang-tidy
+    name: Build and test
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: false
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
+        crypto: [openssl_1.1, openssl_3, boringssl]
```
glhewett, line 38: 👍

.github/workflows/main_ci.yml
```diff
@@ -34,226 +28,116 @@ jobs:
...
     if: github.event.pull_request.draft == false
-    needs: quick-linux-interop-check
-    name: Build for older MacOS
-    runs-on: macos-latest
+    needs: build-and-unit-test
+    name: Build with clang-tidy
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        crypto: [openssl_1.1, openssl_3, boringssl]
```
glhewett, line 123: 👍

CMakeLists.txt
```diff
@@ -44,28 +44,24 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILER_ID MATCHES "GNU"
...
   add_compile_options(/W4 /WX)
+  add_definitions(-DWINDOWS)
 
   # MSVC helpfully recommends safer equivalents for things like
   # getenv, but they are not portable.
   add_definitions(-D_CRT_SECURE_NO_WARNINGS)  
 endif()
 
 if (SANITIZERS)
+  message("Enabling sanitizers")
```
glhewett, line 55:
```suggestion
  message(STATUS "Enabling sanitizers")
```

lib/bytes/test/bytes.cpp
```diff
@@ -1,5 +1,6 @@
 #include <bytes/bytes.h>
 #include <doctest/doctest.h>
+#include <iostream>
```
glhewett, line 3: did you mean to keep this here?

glhewett (approved):
No show stoppers (so approved), but I have some comments.

.github/actions/build/action.yml
bifurcation: Yeah, good point.  I think we are seeing the crypto libraries get rebuilt, thus causing unnecessary slowness.  Two alternatives come to mind:

1. Including the crypto library in the cache key (as you suggest)
2. Somehow filling all three libraries into the cache

The latter approach seems appealing to me, because you would get more sharing across builds for the same platform.  But it would require some refactoring in how we select the crypto library in CMakeLists.txt.

So in the short run, I think the right fix is (1).  I can make `matrix.os` and `matrix.crypto` inputs to this job.

.github/workflows/main_ci.yml
bifurcation: Good point, will do.

PR #409, opened 2023-11-17
Build with vcpkg and OpenSSL 1.1 by default

bifurcation:
Right now, an unadorned `make` grabs whatever OpenSSL the system happens to provide via FindPackage.  This PR changes the behavior of `make` so that it uses vcpkg with OpenSSL 1.1.


glhewett (approved):
LGTM

PR #411, opened 2023-11-17
add support for testing openssl 3.1

glhewett:


bifurcation:
@glhewett Should we just upgrade `alternatives/openssl_3` to 3.1?

glhewett:
> @glhewett Should we just upgrade `alternatives/openssl_3` to 3.1?

I am not sure what adoption looks like, but I would assume that most would be running 3.1 instead of 3.0.  I think upgrading to 3 might be best.

glhewett:
You can close this, and I might consider upgrading vcpkg to 3.1 in the future.

PR #412, opened 2023-11-27
Expose group ID and allow inspection

bifurcation:
Right now, there is no way for application code to inspect an MLS message prior to handling it.  PrivateMessages are of course encrypted until they are handled, but even PublicMessages hide the API to access the message contents.  This PR makes such inspection possible, via an `unwrap()` method that produces `AuthenticatedContent` and a `handle()` variant that accepts `AuthenticatedContent`:

```
const auto content_auth = state.unwrap(message);
// Verify that content_auth has the right contents for the application context
const auto maybe_next = state.handle(content_auth);
```

As a convenience, we also expose `group_id()` on MLSMessage.

### DIFF
--- a/include/mls/messages.h
+++ b/include/mls/messages.h
@@ -570,6 +570,23 @@ private:
   friend struct PrivateMessage;
 };
 
+struct ValidatedContent
+{
+  public:
+  const AuthenticatedContent& authenticated_content() const;
+
+  friend bool operator==(const ValidatedContent& lhs, const ValidatedContent& rhs);
+
+  private:
+  AuthenticatedContent content_auth;
+
+  ValidatedContent(AuthenticatedContent content_auth_in);
+
+  friend struct PublicMessage;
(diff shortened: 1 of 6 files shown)


suhasHere (approved):

bifurcation:
A note while we're waiting for the CI to get fixed -- Having `handle()` accept `AuthenticatedContent` is sub-optimal, because it forces us to assume that content has been validated.  Instead, we could encapsulate `AuthenticatedContent` in a wrapper type that would attest that the AuthenticatedContent had been validated.  Something like:

``` c++
struct ValidatedContent {
  private:
  // Only allow construction from outside friends that will validate
  AuthenticatedContent content_auth;
  ValidatedContent(AuthenticatedContent content_auth);
  friend struct PrivateMessage;
  friend struct PublicMessage;

  public:
  // Provide const / read-only access to the inner content
  const GroupContent& content() const;
};
```

PR #415, opened 2023-12-14
Disable unit tests on Windows

bifurcation:
This PR disables unit testing in Windows CI runs.  As the comment in the workflow file notes for posterity, we have two conflicting bugs.  A [doctest bug](https://github.com/doctest/doctest/issues/836) discovered through #413 which causes doctest to fail when built with ASAN, and a GitHub Actions bug that causes the tests to hang when _not_ built with ASAN.

Disabling unit tests on Windows is unfortunate, but I don't think it's a huge loss in terms of test fidelity.  The behaviors we're testing are higher-level things that I would expect to be pretty portable, not dependent on anything platform-specific.

The long-term fix here probably entails moving to a new test framework, since it looks like [doctest is essentially unmaintained](https://github.com/doctest/doctest/issues/554).  My understanding is that [Catch2](https://github.com/catchorg/Catch2) is a pretty standard alternative, and appears to be better maintained than doctest.  The syntax seems similar (I think by design), so it may be pretty straightforward to change over.

### DIFF
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -55,9 +55,7 @@ jobs:
 
     - name: Build
       run: |
-        # XXX(RLB): If we do not have SANITIZERS=ON here, the Windows CI builds
-        # hang in the middle of unit testing.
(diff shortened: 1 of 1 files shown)


glhewett (approved):
I agree that this is not an ideal situation.  I know that you are already working on a unit test replacement, so let's keep moving forward.

PR #416, opened 2023-12-14
Convert tests to Catch2

bifurcation:
As discussed in #415, doctest is broken and abandoned.  Catch2 seems to be a more standard and more supported test framework.  This PR converts MLSpp's tests from doctest to Catch2.

* Adjust all the `vcpkg.json` files so that Catch2 is available in all configurations
* Updates CMakeLists.txt in the test files to build/link against Catch2
* Makes some minor fixes in test files to fix errors building against Catch2

And for good measure, I went ahead and re-enabled tests on Windows.  We'll see what happens in CI!

### DIFF
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -15,114 +15,114 @@ env:
   CMAKE_TOOLCHAIN_FILE: ${{ github.workspace }}/vcpkg/scripts/buildsystems/vcpkg.cmake
 
 jobs:
-  formatting-check:
-    name: Formatting Check
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v4
-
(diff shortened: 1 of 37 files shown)


bifurcation:
It turns out that what happens in CI is that the Windows tests still hang without `SANITIZERS` and blow up with it.  The latter appears to be a [known issue](https://github.com/catchorg/Catch2/issues/2507), since 2022.  So I have re-disabled Windows tests.  It looks like we're going to have to address the hang more directly.

I think it's still worthwhile to merge this PR, in order to get on a more supported testing framework.

bifurcation:
Merging based on offline 👍 from @glhewett and @birarda

glhewett:
Did you meant to comment out all of the tests?  I guess not: #418

PR #417, opened 2023-12-14
Fix Windows hang in testing

bifurcation:
Since a recent upgrade to the `windows-latest` image, Windows CI builds have been hanging when running certain tests.  This turns out to be due to an error in the way MSVC implements "empty base optimization", which breaks the `overloaded` technique for ad-hoc polymorphic lambdas -- simply defining one [results in stack corruption](https://developercommunity.visualstudio.com/t/runtime-stack-corruption-using-stdvisit/346200).  While Microsoft claims to have fixed this bug, and GitHub claims to use a version that has the fix, the bug still manifests.  This PR adds a work-around that ensures that an `overloaded` type always has at least one member.

### DIFF
--- a/.github/actions/prepare-build/action.yml
+++ b/.github/actions/prepare-build/action.yml
@@ -23,9 +23,9 @@ runs:
       uses: actions/cache@v3
       with:
         path: ${{ inputs.cache-dir }}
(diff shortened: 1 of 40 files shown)


bifurcation:
OK, I think the mystery is resolved here.  I will work on getting the Catch2 change merged, then land this one.

glhewett (approved):
LGTM

PR #424, opened 2024-05-23
Support custom signing routines in SignaturePrivateKey

rcombs:
This enables support for signing with non-extractable private keys (e.g. TPM-bound).

### DIFF
--- a/include/mls/crypto.h
+++ b/include/mls/crypto.h
@@ -240,11 +240,14 @@ struct PublicJWK
 
 struct SignaturePrivateKey
 {
+  using SignerFunc = std::function<bytes(const std::vector<uint8_t>&)>;
+
   static SignaturePrivateKey generate(CipherSuite suite);
   static SignaturePrivateKey parse(CipherSuite suite, const bytes& data);
   static SignaturePrivateKey derive(CipherSuite suite, const bytes& secret);
   static SignaturePrivateKey from_jwk(CipherSuite suite,
                                       const std::string& json_str);
+  static SignaturePrivateKey from_func(SignerFunc func, bytes pub_data);
 
   SignaturePrivateKey() = default;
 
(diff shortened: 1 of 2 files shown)


bifurcation:
Hi @rcombs, thanks for submitting this.  Just to confirm I understand the architecture correctly: 

* The caller would represent a private key as a `SignerFunc`, and then use `SignaturePrivateKey::from_func` to construct the object they provide to MLSpp
* A private key constructed in this way would have its `_sign_func` member populated, but its `priv_data` empty.

I would prefer to handle this a little more systematically, in the sense of not having an alternative path for functions like this as opposed to other keys.  For example, we could have signing always be a function, and supply a default function when constructing a key from bytes.  In line with #425, it would also be good to push most of the complexity here into the `hpke` module.  The structs in `crypto.cpp` are designed to be pretty thin wrappers around the crypto types.

Given those considerations, how about an approach something like the following:

* Move the `sign()` method from `Signature` to `Signature::PrivateKey` 
* Change `SignaturePrivateKey` to hold a `std::unique_ptr<Signature::PrivateKey>` (Honestly I don't remember why we store the bytes right now.  I think maybe pure laziness.)
* ... and have the existing constructors build that pointer as appropriate
* Then add a new constructor that directly provides a pointer to `Signature::PrivateKey`

From the caller's perspective, you would just from `Signature::PrivateKey` instead of implementing `SignerFunc`.

The main drawback that occurs to me is that we end up exposing what are currently some internal interfaces.  But we're going to have to expose that sort of thing anyway if we're going to allow this sort of runtime adaptation.

bifurcation (approved):

bifurcation:
Picking this back up -- This looks fine to me for what it does.  It needs tests, and I would prefer that serialization failed instead of succeeded with invalid data.  But I'll add that in a follow-on.

bifurcation:
Follow-on is #462

PR #430, opened 2024-08-20
Fix invalid parent hash with empty root left subtree

birarda:
Fixes a bug with parent hash generation when the left subtree of the root node is empty.

If left subtree of the root node becomes empty (for example, when first two members are removed by a third group member) the parent hash computed during `TreeKEMPublicKey::update` will be invalid. A subsequent welcome will also have an invalid tree, and the welcome is unprocessable.

This occurred because `parent_hashes` assumed that the root node was always present in the filtered direct path, and therefore used the root node as the initial parent when doing a top-down generation of parent hashes.

### DIFF
--- a/test/state.cpp
+++ b/test/state.cpp
@@ -1107,11 +1107,10 @@ TEST_CASE_METHOD(StateTest, "Parent Hash with Empty Left Subtree")
     states[0].remove_proposal(LeafIndex{ 1 }),
   };
 
-    auto [commit2, welcome2, new_state_2] =
-      states[2].commit(fresh_secret(), CommitOpts{ { removes }, true, false, {} }, {});
-    silence_unused(commit2);
-    states[2] = new_state_2;
-
+  auto [commit2, welcome2, new_state_2] = states[2].commit(
+    fresh_secret(), CommitOpts{ { removes }, true, false, {} }, {});
+  silence_unused(commit2);
+  states[2] = new_state_2;
 
   // Member @2 should have a valid tree, even though its filtered direct path no
   // longer goes to the root.
--- a/test/state.cpp
+++ b/test/state.cpp
@@ -1118,6 +1118,8 @@ TEST_CASE_METHOD(StateTest, "Invalid Tree2")
                        std::nullopt,
                        {} });
 
+  REQUIRE(states[2].tree().parent_hash_valid());
+
   // Member @2 removes member @1
   auto remove_proposal_1 = states[2].remove_proposal(LeafIndex { 1 });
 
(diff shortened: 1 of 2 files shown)


test/state.cpp
```diff
@@ -1069,3 +1069,50 @@ TEST_CASE_METHOD(RelatedGroupTest, "Reinitialize the group")
...
+
+  for (size_t i = 1; i < 4; i++) {
+    states.push_back({ init_privs[i],
+                       leaf_privs[i],
+                       identity_privs[i],
+                       key_packages[i],
+                       welcome0,
+                       std::nullopt,
+                       {} });
+  }
```
bifurcation, line 1102: Since you're only going to use `states[2]` below this line, you could just initialize it directly.
```suggestion
  const auto state_2 = State{
    init_privs[i],
    leaf_privs[i],
    identity_privs[i],
    key_packages[i],
    welcome0,
    std::nullopt,
    {},
  };
```

Similarly, you don't actually need the `states` vector above, you could just define `state_0`.

test/state.cpp
```diff
@@ -1069,3 +1069,50 @@ TEST_CASE_METHOD(RelatedGroupTest, "Reinitialize the group")
...
+
+  // Member @2 removes the members on the left side of the tree
+  const auto removes = std::vector{
+    states[0].remove_proposal(LeafIndex{ 0 }),
+    states[0].remove_proposal(LeafIndex{ 1 }),
```
bifurcation, line 1107: These should be done by member 2

bifurcation:
Thanks for finding this.  Just to recap / restate the issue here:

The problem arises when the whole left subtree of the ratchet tree is empty:

```
      _ = G
      |
    .-+-.    
   /     \   
  _       F
 / \     / \ 
_   _   C   D
```

This can arise, for example, via the following sequence of events:

* A creates a group and adds B, C, D
* C removes A and B

In this case, right now, the algorithms for creating and verifying parent hashes disagree.  On the creation side, the `parent_hashes` function [assumes that the top node of the filtered direct path is the root](https://github.com/cisco/mlspp/blob/8a010ef9ce9f1bdd8b79a11ee1c089440a3f63fb/src/treekem.cpp#L917), so the top parent-hash link it forms is from the penultimate node to the root -- not to its actual parent.  When someone goes to verify this tree, they will find that the top-level node in the committer's direct path does not have a valid parent-hash link to either of its children, since the link was instead built to the root, and thus that the tree is parent-hash invalid.

In the above example tree, `C.parent_hash` would be the parent hash over `G` and the blank left subtree.  It should be over `F` and the (one-node) subtree at `D`.

bifurcation (approved):
Couple of minor simplifications to the test case, but otherwise LGTM.

It seems like we could actually do this as a test case on `TreeKEMPublicKey::update`, but we can do that in a follow-on.

PR #432, opened 2024-09-06
Use volatile stores in bytes destructor

rcombs:
Ensures that the zeroization can never be optimized out.

### DIFF
--- a/lib/bytes/include/bytes/bytes.h
+++ b/lib/bytes/include/bytes/bytes.h
@@ -17,7 +17,11 @@ struct bytes
   bytes& operator=(bytes&&) = default;
 
   // Zeroize on drop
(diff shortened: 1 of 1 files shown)


rcombs:
There are also non-openssl routines that can be used for this (e.g. `memset_explicit`, `memset_s`, `explicit_bzero`, `SecureZeroMemory`…), or `std::fill` could be used with the pointers casted to `volatile`.

rcombs:
Switched this over to `std::fill` with `volatile` pointers, which has the same property of being guaranteed not to be optimized out (since volatile stores cannot be eliminated per the as-if rule), without expanding the openssl dependency; this should also fix the build errors created by including the openssl header.

bifurcation:
Thanks @rcombs that sounds like a nicer solution to me.  Happy to merge once CI passes.

PR #436, opened 2024-11-01
Add support for Light MLS

bifurcation:
As described in [draft-kiefer-mls-light-01](https://www.ietf.org/archive/id/draft-kiefer-mls-light-01.html).

This PR enables clients to join and participate in groups without having the full ratchet tree.  There are a few major changes:

1. The storage for nodes in TreeKEMPublicKey is changed from an array to a map.  Actually two maps, one storing node values and one storing node hashes.
2. AnnotatedWelcome and AnnotatedCommit structs are added, with handlers on State to process them
3. Some changes in TreeKEM logic to support working without the whole tree
4. An `mls_ds` library is added for tools that are useful in constructing a DS.  In particular, a class that follows the evolution of a group's ratchet tree based on the Commits in the group


suhasHere (approved):
Looks good to me. I have not deeply verified the tree math though but others flows looked fine. Thanks

cmd/interop/src/mls_client_impl.cpp
```diff
@@ -1062,14 +1062,16 @@ MLSClientImpl::commit(CachedState& entry,
     by_value.emplace_back(std::move(proposal));
   }
 
-  auto force_path = request->force_path();
-  auto inline_tree = !request->external_tree();
+  const auto force_path = request->force_path();
+  const auto inline_tree = !request->external_tree();
+  const auto include_membership_proof = false;
```
bifurcation, line 1067: Looking at this, we haven't updated this code since we moved to AnnotatedWelcome.  I'll update and request re-review.

suhasHere (approved):

PR #439, opened 2025-01-23
Configure namespace.h into cmake binary dir instead of source

Atlante45:
When calling `configure_file()` on `namespace.h.in`, cmake will create a temporary file at the target location called `namespace.h.tmp` and when done generating move it to `namespace.h`

This causes an issue when building multiple release types at the same time. They will both generate the same .tmp file and overwrite each other. The first that completes will move the shared file causing the other instance to fail with the error `No such file or directory`

This is a proposed solution that generates the namespace header in the binary directory as intended by `configure_file` and then makes sure all relevant steps load the header from the appropriate directory


bifurcation (approved):
Thanks for the PR @Atlante45, and sorry for the very delayed review.

FYI, I filed #448 with a couple of minor follow-ups, so as not to delay merging this.

bifurcation:
I'm not sure why CI is not running on this.  Merging on faith, and verifying later.

PR #458, opened 2026-01-30
Addition of mls::State serialization using the existing serialization options

BlehPoster:
Adds TLS_SERIALIZABLE recursively to all required nested structures while keeping member accessibility untouched by adding specific overloaded stream operators.
Known issue: Since serialization requires an empty/existing object, mls::State and its nested structures must be default constructible. As long as serialization is implemented outside of structures, default constructors are mandatory.
Added missing #ifdef(WITH_PQ)

The project builds locally (Linux, Clang 18), and all unit tests are passing.

### DIFF
--- a/include/mls/crypto.h
+++ b/include/mls/crypto.h
@@ -31,6 +31,8 @@ struct KeyAndNonce
 {
   bytes key;
   bytes nonce;
+
+  TLS_SERIALIZABLE(key, nonce)
 };
 
 // opaque HashReference<V>;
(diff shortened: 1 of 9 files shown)


kosmakoff:
@BlehPoster I tried your code and it was not deserializing the `State` correctly - it did not restore public keys, and because of that the restored State failed to execute `handle()` for example.

I made the fix, if you're interested: https://github.com/kosmakoff/mlspp/commit/95c574faa9b6f8e1463e914a1f8fa9000a02f66f

BlehPoster:
Hi @kosmakoff, sorry for the late reply, I will take a look at this.
But since this is stale for so long, I will close this and search for different solutions.

Issue #459, opened 2026-02-14
Add soname to shared library


bkmgit:
Enables easy identification of abi changes.

PR #460, opened 2026-02-14
Add soname to shared libraries (#459)

bkmgit:


bifurcation:
This looks odd (`#ifdef` in CMakeLists.txt doesn't do anything; it's just a comment).  And we don't really have a versioning strategy for this library.  So I'm going to reject this for now.

bkmgit:
Thanks for information about the `#ifdef`.  A versioning strategy would be helpful for redistribution, but understand if it is not a priority. Have used what was in the PR at https://packages.fedoraproject.org/pkgs/mlspp/mlspp/ but if something different is chosen, can update it.

PR #461, opened 2026-04-09
Add support for non-exportable signature private keys

bifurcation:
## Summary

- Add `ExternalPrivateKey` abstraction in HPKE layer for non-exportable keys
- Support OpenSSL 3.x OSSL_STORE API for loading keys from URIs (pkcs11:, file:, etc.)
- Support OpenSSL 1.1.x ENGINE API for external key providers
- Support callback-based signing for BoringSSL and custom secure enclave integrations
- Add `SignaturePrivateKey::from_external()` and `from_external_callback()` factory methods
- Add `SignaturePrivateKey::exportable()` to check if private key data can be extracted
- Remove TLS serialization from SignaturePrivateKey (non-exportable keys cannot be serialized)

## Test plan

- [x] Existing crypto tests pass (69/69)
- [x] Callback-based external signing test works for all cipher suites
- [x] Invalid URI handling tests for OpenSSL 3.x, OpenSSL 1.1.x, and BoringSSL
- [x] File URI test for OpenSSL 3.x (with optional MLSPP_TEST_KEY_FILE env var)
- [x] macOS Keychain test using Security.framework SecKey APIs

## Notes

- PKCS#11 integration is left as an exercise for integrators (requires pkcs11-provider or SoftHSM setup)
- The macOS Keychain test creates a temporary in-memory key (not persisted to Keychain) for test isolation

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/include/mls/crypto.h
+++ b/include/mls/crypto.h
@@ -356,7 +356,8 @@ public:
 
   // Move operations
   SignaturePrivateKey(SignaturePrivateKey&& other) noexcept = default;
-  SignaturePrivateKey& operator=(SignaturePrivateKey&& other) noexcept = default;
+  SignaturePrivateKey& operator=(SignaturePrivateKey&& other) noexcept =
+    default;
 };
 
 } // namespace MLS_NAMESPACE
--- a/include/mls/crypto.h
+++ b/include/mls/crypto.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <functional>
 #include <hpke/digest.h>
 #include <hpke/hpke.h>
 #include <hpke/random.h>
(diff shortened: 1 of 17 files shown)


include/mls/crypto.h
```diff
@@ -254,28 +255,83 @@ struct PublicJWK
...
+  /// This is useful for BoringSSL and custom secure enclave integrations.
+  /// The callback receives data to sign and returns the signature.
+  static SignaturePrivateKey from_external_callback(
+    CipherSuite suite,
+    SignaturePublicKey pub_key,
+    std::function<bytes(const bytes&)> sign_callback);
+
   SignaturePrivateKey() = default;
 
+  /// Raw private key data (empty for non-exportable keys)
```
bifurcation: This should be `optional<bytes>` to properly reflect optionality.

include/mls/crypto.h
```diff
@@ -254,28 +255,83 @@ struct PublicJWK
...
   void set_public_key(CipherSuite suite);
+
+  /// Export to JWK format (throws if !exportable())
   std::string to_jwk(CipherSuite suite) const;
 
-  TLS_SERIALIZABLE(data)
+  // Note: SignaturePrivateKey is intentionally NOT TLS serializable.
+  // Non-exportable keys cannot be serialized, and applications should
+  // handle key persistence separately (e.g., storing key URIs).
+
```
bifurcation: We could define serialization/deserialization operators (`<<` `>>`) that throw if the key is not serializable.

include/mls/crypto.h
bifurcation: Also, with the change to not being serializable, I think this field can become private.

include/mls/crypto.h
```diff
@@ -254,28 +255,83 @@ struct PublicJWK
...
+  friend bool operator!=(const SignaturePrivateKey& lhs,
+                         const SignaturePrivateKey& rhs)
+  {
+    return !(lhs == rhs);
+  }
 
 private:
+  /// Handle to external (possibly non-exportable) key
+  std::shared_ptr<hpke::Signature::ExternalPrivateKey> external_key_;
```
bifurcation: Should be unique_ptr instead of shared_ptr.

include/mls/crypto.h
```diff
@@ -254,28 +255,83 @@ struct PublicJWK
...
+
   SignaturePrivateKey(bytes priv_data, bytes pub_data);
+  SignaturePrivateKey(
+    std::shared_ptr<hpke::Signature::ExternalPrivateKey> external_key,
+    bytes pub_data);
```
bifurcation:
```suggestion
  SignaturePrivateKey(
    std::shared_ptr<hpke::Signature::ExternalPrivateKey> external_key);
```

Just call `external_key->public_key()`.

bifurcation:
Closing in favor of #424 + #462

